PR #1, opened 2022-10-12
Initial package structure

mortenengen:
This is a suggestion for a structure of the package. Feel free to comment, cheer, disagree, and suggest changes 😄 

Have a look at `structuralcodes.code`. This is where we collect our design codes. Currently this is set up as a factory. With this setup we can signal to contributors what is the expected content of a design code, and we arrange for all design codes being defined in a similar manner.

The code that has som implementation is _fib_ Model Code 2010. Have a look at `structuralcodes.code.mc2010`. The main class of this design code, `MC2010`, inherits from `structuralcodes.core.base.DesignCode`. Having the `DesignCode` abstract base class opens for defining a common interface for all design codes in the future. Furthermore, there is a builder, `MC2010Builder`, and this builder is registered in the `code_factory` in `structuralcodes.code._code_factory`.

Have a look at the implementation of `MC2010`. I have suggested to implement all methods in separate modules, as is shown in the `_concrete_material_properties` module. The functions are imported in a regular manner, and set as attributes in the constructor of the design code.

I think this opens for a good code structure, where the design code class implementation is rather short, and the implementation of the functions can be put in separate modules with different topics. One drawback is that the intellisense in e.g. VS Code is not working properly in this case. This means that if you instantiate the design code, `code = MC2010()` and start typing `code.fcm()`, VS Code will not show a popup with code hints. It would be nice if we could find a way to fix this. However, the method works as expected, and typing `help(code.fcm)` prints the docstring as expected.

The function `set_design_code()` defined in `structuralcodes.code` calls the code factory and sets a global design code that will act as the default code for subsequent calls. There is also a `set_national_annex()` which is not yet implemented, and a `get_design_codes()` which lists the available codes that are registered at the code factory.

Let us go back to `structuralcodes.core` and look at the `concrete` module. This module contains the `Concrete` class which inherits from `structuralcodes.core.base.Material`. The constructor takes an optional design code, and if this is not given, it will use the global. Furthermore, the module defines a tuple with methods that are required to be implemented in the design code, and these are set as attributes on the `Concrete` class by the constructor. Again, perhaps a nice code structure, but the intellisense is still broken.

One thing to discuss is whether the `fcm` and `fctm` methods should be methods or properties of the `Concrete` class. I do not have a strong opinion here. Implementing e.g. the Young's modulus as a property could however be challenging, unless we provide e.g. the aggregate type to the constructor. 

In the future, we could implement other materials, and even sections, in a similar manner.

To start testing, try typing something like:

```python
import structuralcodes

structuralcodes.set_design_code('mc2010')
concrete = structuralcodes.Concrete()

fcm = concrete.fcm(45)
```


mortenengen:
Note that I have set up a github action that runs Black, flake8 and pylint to make sure that we have some objective guidance on our code. We should not merge changes that do not pass these tests. In the future we should also add testing to this pipeline.

DanielGMorenaFhecor:
I just found that if you import the `_concrete_material_properties` functions inside the `MC2010()` class, then autocomplete works perfectly. Otherwise, adding the attributes dynamically with `setattrs()` only will resolve method names at runtime, so we will be only able to use autocomplete in other environments such as Notebooks, once the variable is initialized (or we could explore stub generation). `MC2010()` would look as the following lines:

```
from structuralcodes.core.base import ConcreteDesignCode


class MC2010(ConcreteDesignCode):
    from ._concrete_material_properties import fcm, fctm, fctkmin, fctkmax, Gf, fcd
```

With this modification, Pylint will show some warnings because imports are not being used at the top level, so it may be a good idea to add to the `.pylintrc` file the following attribute:
```
disable = disable-outside-toplevel
```

Also, to make it work, in `_concrete_material_properties.py` it is necessary to add a first argument to all functions so it refers to the instance of the current `MC2010()` class. As an example:

```
def fcm(self, fck: float, delta_f: float = 8.0) -> float:
    return fck + delta_f
```

We cannot set the type of the `self` variable to `self : MC2010` because it would create a circular dependency.

I do not know if this looks pythonic or not, but it works.
¿What do you think about this?

I can propose a draft for review if you grant me write access to this repository. Otherwise, I can just create a fork.

mortenengen:
Hi @DanielGMorenaFhecor, good to have you on board. Thanks for your comment, and thanks for looking into this 😄 I wanted to respond earlier, but had to digest this and other options a bit.

Importing the functions inside `MC2010` kind of solves it, and disabling the necessary checks in Pylint to have it pass our pipeline is fine. However, I am a bit worried about forcing all contributors to add a `self` in the function signature. So, you solved it, but I don't think we should do it this way 😐 By the way, I like your idea of stub generation. However, it also adds another layer of stuff to manage. I hope you do not find this too stubborn and reluctant, but realize that I appreciate the options that we investigate 😏 

With this said, I think we might have over-complicated stuff slightly in the present draft. Perhaps a cleaner structure could be to collect all the functions from the design codes in sub-packages, and expand from there. Let's explore this idea briefly.

In this structure, and for Model Code 2010 specifically, `code.mc2010.__init__.py` would import all available functions from `_concrete_material_properties.py` and other modules. Functions are implemented and documented once, i.e. in these modules.

This sub-package, `code.mc2010`, could serve as basis for the design code specific higher level classes for material, section and element that we have discussed earlier. The sub-package could also be the main entry point for the end-users that do not wish to use the higher level classes, but instead e.g. work with validation of code functions, or use the functions for their own programs or higher level classes.

Now, Sphinx is happy, because we can autodocument all our basic functions, intellisense is happy, because we did not implement it too fancy, and our contributors are happy, so the PRs keep pouring in 🚀 

I do not have a clear idea of how to incorporate the national annexes in this structure, but I do not see it as a deal-breaker at all.

The material class, e.g. `Concrete MC2010`, could behave like advanced dataclasses. Pretty much like in your [example repo](https://github.com/DanielGMorenaFhecor/ejemplo_python), only using the handy `@property` decorator.

I expect that the design code sub-package and the related material class will have an API which is slightly dictated by the design code. And this API will therefore be different for the various design codes that will be implemented in the long run. Which I think is a good approach. However, in the case of the higher level classes for sections and elements, the public API should not vary that much, and designing common interfaces for these using `ABC`s sounds very promising. E.g. all `ReinforcedConcreteRectangular` should have a `moment_curvature` method, irrespective of which design code it is based on, however the implementation of course changes.

We could even implement factories for the higher level classes similar to the `SimpleConcreteFactory` in your example repo, or the `DesignCodeFactory` in the present example.

@talledodiego and @DanielGMorenaFhecor, do you see any immediate issues with this structure?

DanielGMorenaFhecor:
Hi @mortenengen amd @talledodiego! I have been checking your new proposed changes. 

So, as a recap, our idea is to use `mc2010.__init__.py` to import all function definitions from the desired modules. Just as an example, at the beginning of this file, we would write `from ._concrete_material_properties import *` . For me, it is OK, as they are going to be stateless functions that do not need to keep track of any object!

Just as a note, using the wildcard character `*` for the import statements is not usually a recommended practice, but in this case, I do not see any problems. Let me know if you think the opposite.

I have been thinking about how to deal with national annexes. One possible approach could be to create a new sub_module under de mc2010 root that takes care of the national annex (e.g. `mc2010._national_annex.py`) that contains the next definitions:

```
import typing as t

_NATIONAL_ANNEX: t.Optional[str] = None

def set_national_annex(national_annex='default') -> None:
    global _NATIONAL_ANNEX
    _NATIONAL_ANNEX = national_annex

def get_national_annex() -> str:
    return _NATIONAL_ANNEX
```

This way, we have a variable `_NATIONAL_ANNEX` we can point to through imports in other local modules. If we also import this file in the `mc2010.__init__.py` file, we allow the user to externally set up the desired national annex as stands in the following workflow, which I find pretty intuitive:

```
import structuralcodes.code.mc2010 as mc2010

mc2010.set_national_annex('ES')
fck = mc2010.fcm(fck=25, delta_f=8)
print(f'fck={fck}')

annex = mc2010.get_national_annex()
print(f'National Annex={annex}')
```

With respect to using the material, section, and element base classes, let me know if I understood properly: the idea is to use the structure from `code.mc2010` (e.g. symbols, properties, ...) to define the main properties of these abstract entities (`ABC`) that have not any implementation yet and then define different advanced subclasses like `MC2010Concrete` or `ECRectangularSection`  which currently call to their respective code functions. 
This way, I find it quite useful and straightforward for new contributors!

Anyway, I will communicate these updates to Alejandro, Carlos, and Javier so that they can keep track of the process. In addition, we internally discussed if it would be better to use Eurocode symbols and definitions for the main abstract classes `Concrete`, `Section`, and `Element`, but I think that they can offer more light in this area than I, that I am more code-focused! So we may chat about it later.

Thank you for your time!

mortenengen:
Thanks for the checking, @DanielGMorenaFhecor 😃 

Your recap of the idea behind the design code sub-packages is correct. As a remark, I think we should avoid the wildcard import, and rather import all the functions explicitly. A little bit more typing, but it makes the code so much easier to read afterwards 😃 

I like your idea related to the national annexes, i.e. storing the label of the national annex in a global variable. However, the function `set_national_annex` could perhaps live in a module near the root of the main package, and furthermore live in the root namespace? In this way, the global variable that is set by `set_national_annex` could be the same for all the design codes that the user might want to include in the calculations. On the other hand, the details of the specific national annex of each design code should live in a sub-module under the respective design code, and from there, each design code should implement its own `get_national_annex` function. Does this make sense?

Your workflow would now look like this:

```python
import structuralcodes.code.mc2010 as mc2010

from structuralcodes import set_national_annex

set_national_annex('ES')
fck = mc2010.fcm(fck=25, delta_f=8)
print(f'fck={fck}')

annex = mc2010.get_national_annex()
print(f'National Annex={annex}')
```

I think you got the idea of the materials, sections and elements.

Just let me emphasize that the materials, e.g. `ConcreteMC2010`, will have property names that correspond to the nomenclature in Model Code 2010. Other concretes might have other property names if their nomenclature is different.

On the other hand, on the section and element level, we might implement most of the interface in `ABC`s, since the methods of the sections and elements will most likely have the same names, irrespective of the code they are related to.

Summing up, the `ABC`s for the materials, e.g. `Concrete`, `Reinforcement` etc. will have just a very limited number of abstract stuff implemented, if anything at all. On the other hand, the `ABC`s for sections, e.g. `RectangularSection` and `CircularSection` will have most of the interface implemented as abstract methods. Any thoughts on this?

mortenengen:
Closed with ca0ff0e.

PR #5, opened 2022-11-18
Concrete refactoring dt

talledodiego:
First attempt to refactor concrete:

- one abstract class and then subclasses for each design code
- new implementation on how attributes are used: by default they assume a None value. If overwritten by user, will use the new values
- added optional description/name for materials and concrete. Default behaviors for names are defined
- created a factory function that creates the proper type of concrete based on the code wanted (the active global one or another specified by user)
- simple test.py file shows the use of the new interface

TO BE DONE STILL:
- split the class in more files (for now it is all in a single concrete.py file): one file for abstract class and factory function + one file for each concrete implementation for the different codes
- some more value checking in the setters of properties in concrete
- improve docstrings (sphinx format is ok?)
- add concrete for Mc2020, Eurocode 1992:2004 and ACI once the framework is approved


mortenengen:
Continued in #7

PR #21, opened 2023-03-21
feat: Add constitutive laws

talledodiego:
- add base abstract class of constitutive law objected
- add some class implementation for some common constitutive laws (elastic, elastic-plastic, parabola rectangle and user-defined
- the functions accept a float, a list or a numpy array

TODO:

- [ ] implement more tests
- [x] implement get_secant method


mortenengen:
Continued in #27.

PR #23, opened 2023-05-11
feat: creep and shrinkage according to EN 1992 Annex B

sshaave:
annex_b_shrink_and_creep.py: the python code executing the calculations in Annex B
test_ec2_2004_annex_B_shrink_and_creep.py: the pytest


structuralcodes/codes/ec2_2004/annex_b_shrink_and_creep.py
```diff
@@ -0,0 +1,184 @@
...
+            1.5 * (1 + (0.012 * RH) ** 18) * h_0 + 250 * alpha_3,
+            1500 * alpha_3,
+        )
+        phi_RH = (
+            1 + (1 - RH / 100) / (0.1 * h_0 ** (1 / 3)) * alpha_1
+        ) * alpha_2
+    return beta_H, phi_RH
+
+
+def annex_B_shrinkage(
```
talledodiego, line 104: We should have from annex B equations for B.11 and B.12 (the next function in this module).
The logic in this function should be inside module for chapter 3 and not annex b?
Maybe consider factoring out this function?
Anyhow, again I would split all the small steps:
1. Function for 3.8 eps_cs (eps_cd, eps_ca)
2. Function for eps_cd (3.9) with inputs beta_ds, kh, eps_cd_0 (for this one the user should use the function from Annex B module).
3. Function for beta_ds(t,ts) with inputs t, ts, h0
4. funtion eps_ca with inputs beta_as and eps_ca_inf (eq. 3.11)
5. Function for eps_ca_inf with input fck (eq. 3.12)
6. Functions for beta_as with input t (eq. 3.13)

talledodiego (changes requested):
Thanks for this valuable contribution.
I would tend to refactor a bit the code in order to be more adherent to the equations in the code, as reported in the comments below.
As a note I would tend to factor out the shrinkage computation from chapter 3, using annex b only for computation of eps_cd_0.

Linting reports lines too long. Maybe you can fix them.

Finally test misses 11 lines. You should try to achieve 100% test coverage of the developed modules.

structuralcodes/codes/ec2_2004/annex_b_shrink_and_creep.py
sshaave: Hi. Thanks for many good suggestions to my code. The code was written in a very specific way to yield exactly what I needed to train my AI. Traces of this are still seen and will be adressed (for instance: calculating dead and live load for 7 and 8 days to inf). I hope I will have my code ready for a new review before next week.

sshaave: I have pushed my new code. Should I now press the "ready for review" button?

talledodiego (approved):
Thanks for the review.
The code seems excellent, thanks for the effort!
@mortenengen can we merge this code to dev branch?

PR #50, opened 2023-09-29
Algorithm marin

aperezcaldentey:
First draft of Marin Algorithm for integration of polynomial surfaces on a polygon

PR #75, opened 2023-12-26
Concrete EC2 2023 Implementation

DanielGMorenaFhecor:
I have created a draft for  `Concrete EC22023` implementation. I have marked all points to discuss with `TODO` flags throughout the scripts. 

@aperezcaldentey I have created this in another branch as I feel that this is cleaner than using your current #50 branch. 
I am mentioning @mortenengen  and @talledodiego you both, as I feel we can all discuss this so we can uniform future `Concrete` implementations.

Points to discuss:

- Review the optional args that are passed to the constructor (`__init__`). I have only picked the necessary parameters that are needed for computing properties that REALLY own to the concrete material, like `gamma_C`, `kE` and `concrete_dev_class`
- For all those concrete properties that depend on external non-inherent values (like time dependent properties `t0` or `t_ref`) instead of properties, I have created methods in which we can pass those values (for example `fctd(self, t_ref: float)` for computing the design tensile strength).
- I have not included any property that involves any geometrical parameter (like `phi_50y_t0` for creeping) so I think this should be handle by the top level section class together with the material.

Remaining:

- [x] `Concrete EC22023` test suite.
- [ ] Add proper documentation to all functions (though some docstrings are already written).


structuralcodes/materials/concrete/_concreteEC2_2023.py
```diff
@@ -0,0 +1,194 @@
...
+                raise ValueError(
+                    (
+                        'Mean compressive strength cannot be lower than',
+                        'characteristic strength.\n',
+                        'Current characteristing strength: ',
+                        f'fck = {self._fck}.',
+                        f'Current value: value = {value}',
+                    )
+                )
+            self._reset_attributes()
```
mortenengen, line 99: We should not reset the attributes in the setters. If a user first set e.g. `fcm`, and furthermore `fctm`, the value that was set to `fcm` will be replaced with `None`. This comment also applies to the setters of `fctm` and `Ecm`.

structuralcodes/materials/concrete/_concreteEC2_2023.py
```diff
@@ -0,0 +1,194 @@
...
+    _Ecm = t.Optional[float] = None
+
+    # TODO: do we agree with the optional keyword arguments here?
+    def __init__(
+        self,
+        fck: float,
+        name: t.Optional[str] = None,
+        density: float = 2400.0,
+        kE: float = 9500,
+        concrete_dev_class: str = 'CS',
```
mortenengen, line 30: Could we rename `concrete_dev_class` to `strength_dev_class`? This also applies to the relevant methods in the [materials section](https://github.com/DanielGMorenaFhecor/structuralcodes/blob/238d7d562c01f6e934628bf39cae9a5a4ca43265/structuralcodes/codes/ec2_2023/_section5_materials.py).

structuralcodes/materials/concrete/_concreteEC2_2023.py
```diff
@@ -0,0 +1,194 @@
...
+                t_ref (float): the reference time in days
+                t0 (float): age at loading in days
+                fck_ref (float, optional): the reference compressive strength
+                    MPa (default is 40 MPa).
+
+            Returns:
+                float: the design compressive strength of concrete in MPa
+
+            Raises:
+                ValueError: if fkc_ref is less or equal to 0
```
mortenengen, line 159: Typo: fkc_ref -> fck_ref 😃

mortenengen (changes requested):
Nice contribution, @DanielGMorenaFhecor 👍  I have suggested a couple of fixes. Also, please add tests 😃

mortenengen:
@DanielGMorenaFhecor I have committed a couple of fixes just to make the code pass our pipeline. I hope you are fine with this 😃

structuralcodes/materials/concrete/_concreteEC2_2023.py
talledodiego: I also add: can it be helpful to recognize both CS, CN, CR and slow, normal, rapid?

structuralcodes/materials/concrete/_concreteEC2_2023.py
```diff
@@ -0,0 +1,195 @@
...
+        def fcm(self) -> float:
+            """Returns the mean strength of concrete.
+
+            Returns:
+                float: The mean compressive strength in MPa.
+            """
+            if self._fcm is not None:
+                return self._fcm
+            self._fcm = ec2_2023.fcm(self.fck)
+            return self._fcm
```
talledodiego, line 77: this lines can be simply condensed in one:
`return ec2_2023.fcm(self.fck)`

structuralcodes/materials/concrete/_concreteEC2_2023.py
talledodiego: I agree. reset is called by parent class `Concete` when `fck` is set. Indeed when `fck` is set as a new value is like having a "totally different" concrete, therefore the reset.

structuralcodes/materials/concrete/_concreteEC2_2023.py
```diff
@@ -0,0 +1,195 @@
...
+                return self._Ecm
+            self._Ecm = ec2_2023.Ecm(self.fcm, self._kE)
+            return self._Ecm
+
+        @Ecm.setter
+        def Ecm(self, value: float) -> None:
+            self._reset_attributes()
+            self._Ecm = value
+
+        def fcd(self, t_ref: float, t0: float, fck_ref: float = 40) -> float:
```
talledodiego, line 146: I have a doubt. With this function we can compute design strength of concrete developed during time.
Most of the times I think users just need fcd (28 days). Could't we make the time-related parameter as optional with default values in order to have by default from fcd the design strength at 28 days?

talledodiego (changes requested):
Excellent contribution!

talledodiego:
> * Review the optional args that are passed to the constructor (`__init__`). I have only picked the necessary parameters that are needed for computing properties that REALLY own to the concrete material, like `gamma_C`, `kE` and `concrete_dev_class`
> * For all those concrete properties that depend on external non-inherent values (like time dependent properties `t0` or `t_ref`) instead of properties, I have created methods in which we can pass those values (for example `fctd(self, t_ref: float)` for computing the design tensile strength).
> * I have not included any property that involves any geometrical parameter (like `phi_50y_t0` for creeping) so I think this should be handle by the top level section class together with the material.

I agree with your points.
I really did not think thoroughly about safety factors. We could do as you have done providing in input the gamma for the material. This case is fairly simple because we have just one factor (1.5) for concrete. Even if for more controls in production a lower factor could be considered. For other materials it can be more complicated depending on the structural element / connections etc. Further it may depend on type of actions like for accidental actions where we can use lower gammas. I don't know if it is better to include some sort of partial safety factor inside the code-level and at the material level our classes would ask to that the partial factor to use... what do you think @DanielGMorenaFhecor @mortenengen ? Too much complicated?

mortenengen:
> ... I don't know if it is better to include some sort of partial safety factor inside the code-level and at the material level our classes would ask to that the partial factor to use... what do you think @DanielGMorenaFhecor @mortenengen ? Too much complicated?

I think it makes sense to leave it as in the draft for now, and revisit this topic when we start implementing national annexes.

> ... we could implement here in Concrete class a `set_notional_size(float)` and then from section call `material.set_notional_size(2*A/u)`. Makes sense?

This sounds like an elegant solution. So leave it as is for now, and revisit when this functionality is implemented in the section classes?

structuralcodes/materials/concrete/_concreteEC2_2023.py
DanielGMorenaFhecor: Done!

structuralcodes/materials/concrete/_concreteEC2_2023.py
DanielGMorenaFhecor: Perfect! I have also added the corresponding tests for the _section5_materials.py module

structuralcodes/materials/concrete/_concreteEC2_2023.py
DanielGMorenaFhecor: Fixed!

structuralcodes/materials/concrete/_concreteEC2_2023.py
DanielGMorenaFhecor: Mmm! Just wondering, I guess that we can do something as follows:
```
@property
def fcm(self) -> float:
    """Returns the mean strength of concrete.

    Returns:
        float: The mean compressive strength in MPa.
    """
    self._fcm = self._fcm or ec2_2023.fcm(self.fck)
    return self._fcm
```

I guess that we may like to store the `self._fcm` value just in case that the parent class calls `_reset_attribules(self)`, don't we?

structuralcodes/materials/concrete/_concreteEC2_2023.py
DanielGMorenaFhecor: Done! I have also added the same optional argument for `fctd`

DanielGMorenaFhecor:
### Changes and Modifications:

- Implemented changes based on prior suggestions.
- Enhanced documentation with docstrings and comments.
- Developed tests for the `ConcreteEC2_2023` class.
- Revised the parameter from `concrete_dev_class` to `strength_dev_class`, introduced options "slow," "normal," and "rapid." Adjusted relevant tests in `_section5_materials` to reflect these updates.
- Transitioned the `sigma_c` property to a method to accommodate `eps_c` as a parameter.

### Discussion Points:

- Cached computed properties. This approach caches the results of properties (like for example `_eps_c1`) after their initial invocation. If you are thinking about removing these stored values, I can modify it.
- Do not know if this is the best place to post it but, should we organize test files into distinct directories for better clarity and maintainability? We may have a bunch of files in the near future.

structuralcodes/materials/concrete/_concreteEC2_2023.py
mortenengen: I agree with your suggestion, @DanielGMorenaFhecor

structuralcodes/materials/concrete/_concreteEC2_2023.py
mortenengen: I like that you have changed the signature such that the function represents the equation it implements. This way the user can choose to calculate the arguments with the other implemented functions, or simply provide other values directly 👍

mortenengen:
> ...
>     * Do not know if this is the best place to post it but, should we organize test files into distinct directories for better clarity and maintainability? We may have a bunch of files in the near future.
Good point, I will create an issue for this 👍

mortenengen (approved):

structuralcodes/materials/concrete/_concreteEC2_2023.py
talledodiego: I am not sure I am getting it correctly...
In my original proposal I did the following reasoning:
- by default we have some attributes that have some values based on equations of the code (like fcm).
- the user can always override these defaults.

That's way I did the following:
self._fcm as default is None. The getter return the code result if it is None. If the user override this, the getter will return the user-set value. When we reset is because we want to go back to code-equation computations.
I wasn't thinking at those internal values for caching results.

As far as I can understand the code you are proposing cache the result, so we (and the user) no longer can keep track on what attributes have been override and which have not been override...

I don't know if I am making sense...

talledodiego (approved):

structuralcodes/materials/concrete/_concreteEC2_2023.py
mortenengen: The way this is written in the commited code proposal seems to follow the original reasoning described by @talledodiego above, however written slightly different than in e.g. `ConcreteMC2010` where the value from the code equation is not stored in the internal attribute. I suggest to keep the proposal as it is 😃

mortenengen:
@talledodiego has a good point above, since keeping the internal attribute unset can be a convenient way to signal which attributes are set by the user and which are default values. However, caching the values will improve the speed of the calculations, and we should aim at making this as snappy as possible. Let us keep it like @DanielGMorenaFhecor has proposed 👍

talledodiego:
> @talledodiego has a good point above, since keeping the internal attribute unset can be a convenient way to signal which attributes are set by the user and which are default values. However, caching the values will improve the speed of the calculations, and we should aim at making this as snappy as possible. Let us keep it like @DanielGMorenaFhecor has proposed 👍

perfect, I agree. Caching can speed up for sure the future calls to the same attribute. I suggest anyhow to (later on) add something to keep track on which attributes has been set overwritten by user and which are by-code (i.e. default) computed. In this way we could add a method that returns the list of default and non-default values.

PR #78, opened 2024-03-04
chore: settings.json for Ruff extension in VSCode

DanielGMorenaFhecor:
I have added the following config to the `settings.json` file so we can use the [Ruff Extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff) for VSCode for code autoformatting:

```
    "[python]": {
        "editor.formatOnSave": true, // Enable auto format on save
        "editor.codeActionsOnSave": {
          "source.fixAll": "explicit",
          "source.organizeImports": "explicit" // This lets ruff reorganize the imports
        },
        "editor.defaultFormatter": "charliermarsh.ruff"
    }
````

By default, it reads the config from `pyproject.toml`file.

I just want to let you know three main key points here that we should take into consideration before merging with `dev`:

- I have enabled `formatOnSave` option, for me is useful but do you consider that we should enable this for all contributors?
- I wanted the Ruff formatter to reorganize the import section of each `*.py` file. For me it also makes clearer this section, but we can omit it if you feel so.
- I have also added `exclude = ['tests/*',]` in the `pyproject.toml` file, so ruff does not raise warnings on some double-quoted strings in the tests folder.

Waiting for your response on these 3 keypoints!


mortenengen (approved):

mortenengen:
Thanks for a nice contribution, @DanielGMorenaFhecor! I have accepted the changes, and even removed the excluded test folder and updated all the code so that it passes all tests 😃

PR #82, opened 2024-04-06
chore: group tests in sub-packages

alllexxx1:


alllexxx1:
@mortenengen, you are quite welcome )

PR #83, opened 2024-04-09
feat: add reinforcement material

mortenengen:
Implementation of material classes for reinforcement steel.

## To do
- [x] Add functions for calculating `fyd` for all relevant codes.
- [x] Add `fyd` property for all reinforcement classes.
- [x] Add tests.
- [ ] Add constitutive law property after merging the sections framework branch.

Constitutive law can be added as e.g.:

```python
...
    self._constitutive_law = ...

    @property
    def constitutive_law(self):
        return self._constitutive_law

    @constitutive_law.setter
    def constitutive_law(self, law):
        self._constitutive_law = law
...
```


structuralcodes/materials/reinforcement/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+]
+
+REINFORCEMENTS: t.Dict[str, Reinforcement] = {
+    'fib Model Code 2010': ReinforcementMC2010,
+    'EUROCODE 2 1992-1-1:2004': ReinforcementEC2_2004,
+    'EUROCODE 2 1992-1-1:2023': ReinforcementEC2_2023,
+}
+
+
+def create_reinforcement(
```
talledodiego, line 27: I express the doubt here, but is valid for the constructors of all reinforcement concrete classes:
fyk and E are not the only characteristics needed.
Important parameters e.g. for sectional design is ultimate strain eps_uk and (optionally if you want to include hardening) the tensile strength ftk.

In many codes (EC, MC, etc.) there are expressed the definitions for ductility classes, does it makes sense this idea:
1) the user can input in the constructor (optionally) the values for ultimate strength and strain. 
Alternatively:
2) The user can input the class letter (A, B, C, D) and we compute ftk and eps_uk from the requirements of each class.

In this way for design the user can select a steel 450C for instance as (fyk = 450, class ='C'). For verification (e.g. evaluation of existing structure) the user may have from experimental tests the values and create the material as (fyk = 450, ftk = 600, eps_uk = 0.12)

What do you think @mortenengen ?

talledodiego (commented):
Excellent PR!
I would think about adding as attributes (and relative methods) the tensile strength ftk and the ultimate strain eps_uk.

structuralcodes/materials/reinforcement/__init__.py
talledodiego: Perfect

talledodiego (approved):
excellent job.

PR #86, opened 2024-05-15
feat: handle partial factors for materials

mortenengen:
An attempt to unify the treatment of partial factors between materials and codes. The idea is the following:

- For the `Concrete` class we implement an abstract property `gamma_c`. Similar with `gamma_s` for the `Reinforcement` class.
- `gamma_c` is an optional argument to the constructor of `Concrete`, which sets an internal attribute `_gamma_c`. Similar for the `Reinforcement`.
- The concrete implementation of the `gamma_c` property should either return the value of `_gamma_c` if it is not `None`, or return the value associated with the globally set national annex. In the current implementation, lacking an implementation of the national annexes, the concrete implementations simply return the default values.
- Note that the functions for calculating e.g. `fcd` and `fyd` do not have default values for `gamma_c` or `gamma_s`, and that the related methods on the `Concrete` and `Reinforcement` classes get the value of `gamma_c` and `gamma_s` from the related attribute.

Note that this is similar to the current implementation of `ConcreteEC2_2023`, but with the addition of creating the `property` for `gamma_c` which allows us to connect to the national annexes when the getter is called.

We can also consider using this setup for handling other NDPs.

### DIFF
--- a/structuralcodes/codes/ec2_2023/_section5_materials.py
+++ b/structuralcodes/codes/ec2_2023/_section5_materials.py
@@ -570,7 +570,7 @@ def k_tc(t_ref: float, t0: float, strength_dev_class: str) -> float:
     return 0.85
 
 
-def fcd(fck: float, _eta_cc: float, _k_tc: float, gamma_C: float) -> float:
+def fcd(fck: float, _eta_cc: float, _k_tc: float, gamma_c: float) -> float:
     """Computes the value of the design compressive strength of concrete.
 
     EN 1992-1-1:2023, Eq. (5.3)
@@ -583,7 +583,7 @@ def fcd(fck: float, _eta_cc: float, _k_tc: float, gamma_C: float) -> float:
             member
         _k_tc (float): factor for taking into consideration high
             sustained loads and of time of loading
-        gamma_C (float): partial factor of concrete
+        gamma_c (float): partial factor of concrete
 
     Returns:
         float: the design compressive strength of concrete in MPa
@@ -591,16 +591,16 @@ def fcd(fck: float, _eta_cc: float, _k_tc: float, gamma_C: float) -> float:
     Raises:
         ValueError: if fck is less than 12 MPa
         ValueError if _etc_cc is not between 0 and 1
-        ValueError: if gamma_C is less or equal to 0
+        ValueError: if gamma_c is less or equal to 0
     """
     if fck < 12:
         raise ValueError(f'fck={fck} must be larger or equal than 12 MPa')
     if _eta_cc < 0 or _eta_cc > 1:
         raise ValueError(f'_eta_cc={_eta_cc} must be between 0 and 1')
-    if gamma_C <= 0:
-        raise ValueError(f'gamma_C={gamma_C} must be larger than 0')
+    if gamma_c <= 0:
+        raise ValueError(f'gamma_c={gamma_c} must be larger than 0')
 
-    return _eta_cc * _k_tc * fck / gamma_C
+    return _eta_cc * _k_tc * fck / gamma_c
 
 
 def k_tt(t_ref: float, strength_dev_class: str) -> float:
@@ -642,7 +642,7 @@ def k_tt(t_ref: float, strength_dev_class: str) -> float:
     return 0.7
 
 
-def fctd(_fctk_5: float, _k_tt: float, gamma_C: float) -> float:
+def fctd(_fctk_5: float, _k_tt: float, gamma_c: float) -> float:
     """Computes the value of the design tensile strength of concrete.
 
     EN 1992-1-1:2023, Eq. (5.5)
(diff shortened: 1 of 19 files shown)


structuralcodes/codes/ec2_2023/_section5_materials.py
```diff
@@ -651,21 +651,21 @@ def fctd(_fctk_5: float, _k_tt: float, gamma_C: float) -> float:
...
-        gamma_C (float): partial factor of concrete
+        gamma_c (float): partial factor of concrete
 
     Returns:
         float: the design tensile strength of concrete in MPa
 
     Raises:
         ValueError: if fctk_5 is less than 0
-        ValueError: gamma_C is less than 1
+        ValueError: gamma_c is less than 1
```
talledodiego, line 661: why for fctd, ValueError if gamma_c is less than 1, while for fcd ValueError if gamma_c is less than 0?

structuralcodes/materials/concrete/_concreteEC2_2004.py
```diff
@@ -167,16 +174,22 @@ def Ecm(self, value: float):
         """
         self._Ecm = abs(value)
 
-    def fcd(self, alpha_cc: float, gamma_c: float) -> float:
+    def fcd(self, alpha_cc: float) -> float:
```
talledodiego, line 177: This is making me wonder if we can extend further the concept of partial factors.
For instance also alpha_cc has a recommended value of 1.0 or set by NAs between 0.8 and 1.0.

I think we could:

1. manage also alpha_cc in a way that can interact with NA or return the default value of 1.0

2. Alternatively we could simply se a default value to alpha_cc parameter in this function?

structuralcodes/materials/concrete/_concreteMC2010.py
```diff
@@ -167,3 +173,21 @@ def Gf(self, value: float):
             value (float): the value of Gf in N/m
         """
         self._Gf = abs(value)
+
+    @property
+    def gamma_c(self) -> float:
+        """The partial factor for concrete."""
+        return self._gamma_c or 1.5
+
+    def fcd(self, alpha_cc: float) -> float:
```
talledodiego, line 182: should alpha_cc have a default value of 1.0? (see 7.2.3.1.4)

structuralcodes/materials/concrete/_concreteMC2010.py
```diff
@@ -167,3 +173,21 @@ def Gf(self, value: float):
...
+        """Calculate the design compressive strength.
+
+        Args:
+            alpha_cc (float): A factor for considering long-term effects on the
+                strength, and effects that arise from the way the load is
+                applied.
```
talledodiego, line 188: eventually add default value)

structuralcodes/materials/reinforcement/_reinforcementEC2_2004.py
```diff
@@ -27,6 +28,7 @@ def __init__(
             ftk (float): Characteristic ultimate strength in MPa.
             epsuk (float): The characteristik strain at the ultimate stress
                 level.
+            gamma_s (Optional(float)): The partial factor for reinforcement.
```
talledodiego, line 31: add default value?

structuralcodes/materials/reinforcement/_reinforcementEC2_2023.py
```diff
@@ -27,6 +28,7 @@ def __init__(
             ftk (float): Characteristic ultimate strength in MPa.
             epsuk (float): The characteristik strain at the ultimate stress
                 level.
+            gamma_s (Optional(float)): The partial factor for reinforcement.
```
talledodiego, line 31: add default value?

structuralcodes/materials/reinforcement/_reinforcementMC2010.py
```diff
@@ -27,6 +28,7 @@ def __init__(
             ftk (float): Characteristic ultimate strength in MPa.
             epsuk (float): The characteristik strain at the ultimate stress
                 level.
+            gamma_s (Optional(float)): The partial factor for reinforcement.
```
talledodiego, line 31: add default value?

talledodiego (approved):
Excellent! Just minor notes for you to see

structuralcodes/materials/concrete/_concreteEC2_2004.py
mortenengen: Yes, I am thinking the same. But in my opinion, changing also alpha_cc and others according to the approach suggested here form gamma_c is a next step after we agree on the present suggestion 😃

structuralcodes/codes/ec2_2023/_section5_materials.py
mortenengen: Good point. This was implemented in #24. I have updated it in a53611c.

mortenengen: And I updated similarly for `gamma_s` in bde6d60.

structuralcodes/materials/concrete/_concreteMC2010.py
mortenengen: Fixed in b5f0e82

structuralcodes/materials/concrete/_concreteMC2010.py
mortenengen: Fixed in b5f0e82.

structuralcodes/materials/reinforcement/_reinforcementEC2_2004.py
mortenengen: Fixed in 3a09e31.

structuralcodes/materials/reinforcement/_reinforcementEC2_2023.py
mortenengen: Fixed in 3a09e31.

structuralcodes/materials/reinforcement/_reinforcementMC2010.py
mortenengen: Fixed in 3a09e31.

talledodiego:
Perfect! I think everything looks fine. We can merge it!

PR #87, opened 2024-05-23
feat: use design strengths in constitutive laws

mortenengen:
By default, we should use the design material strengths in the constitutive laws for our material classes. This is a suggestion for how to deal with that.

### DIFF
--- a/structuralcodes/codes/ec2_2004/__init__.py
+++ b/structuralcodes/codes/ec2_2004/__init__.py
@@ -18,6 +18,7 @@ from ._concrete_material_properties import (
     n_parabolic_rectangular,
 )
 from ._reinforcement_material_properties import (
+    epsud,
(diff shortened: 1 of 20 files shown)


talledodiego:
@mortenengen I have done an attempt of complying with everything you asked:
- we have fyd for all Reinforcement classes and the default constitutive laws are based on design values
- I removed Sargin from concreteEC2_2023: I maintained anyhow the properties eps_c1 and eps_cu1, do you agree? Based on this, for analogy we should add later on also the other strains (eps_c2, eps_c3 etc.) needed for defining parabola-rectangle, linear-plastic, stress block etc. We should do the same for other two concrete implementations.

mortenengen:
Very good, and thanks for the additions, @talledodiego 😃

PR #88, opened 2024-05-26
feat: Add Sargin Constitutive Law to Constitutive laws

talledodiego:
Sargin Constitutive Law is added to the list of Constitutive laws. 
Methods for getting stress and tangent are implemented.

Method for getting Marin coefficients is not yet implemented.
The problem is that it seems impossible to represent Sargin as a polynomial function.

My proposal is the following:
- in the `__marin__` method convert the function to a piecewise linear function using `UserDefined`. Then use `__marin__` from UserDefined. 
- a further possibility is that the default `__marin__` for base class `ConstitutiveLaw` those this use of `UserDefined` law for defining a piecewise linear function, so all constitutive laws that are implemented from now on will have a free `__marin__` function for estimating coefficients. If one can and wants, it is possible to define `__marin__` for the derived class to obtain a more performant version (e.g. as we have done for parabola-rectangle end elastic-plastic laws)

### DIFF
--- a/structuralcodes/core/base.py
+++ b/structuralcodes/core/base.py
@@ -4,6 +4,8 @@ import abc
 import typing as t
 import warnings
 
+import numpy as np
+
 import structuralcodes.core._section_results as s_res
 
 
@@ -102,9 +104,50 @@ class ConstitutiveLaw(abc.ABC):
     def __marin__(self, **kwargs):
         """Function for getting the strain limits and coefficients
         for marin integration.
+
+        By default the law is discretized as a piecewise linear
+        function. Then marin coefficients are computed based on this
+        discretization.
         """
-        del kwargs
-        raise NotImplementedError
+
+        # Discretize the constitutive law in a "smart way"
+        def find_x_lim(x, y):
+            # Find the index where x changes from negative to non-negative
+            zero_crossing_index = np.argmax(x >= 0)
+
+            # Check if there are non-zero values for x > 0
+            if np.any(y[zero_crossing_index:] != 0):
+                # Find the last non-zero index for x > 0
+                non_zero_indices = np.nonzero(y[zero_crossing_index:])[0]
+                x_lim_index = zero_crossing_index + non_zero_indices[-1]
+                return x[x_lim_index]
+            # All values are zero for x > 0
+            return None
+
+        eps_max, eps_min = self.get_ultimate_strain()
+        if eps_max > 1:
+            eps_max = 1
+        eps = np.linspace(eps_min, eps_max, 100000)
+        sig = self.get_stress(eps)
+        sig[(sig > 0) & (sig < np.max(sig) * 1e-6)] = 0
+        eps_lim = find_x_lim(eps, sig)
+        # Now discretize the function in 10 steps for positive and 10 steps for
+        # negative part
+        if eps_lim is None:
+            eps = np.linspace(eps_min, 0, 10)
+        else:
+            eps = np.concatenate(
+                (
+                    np.linspace(eps_min, 0, 10, endpoint=False),
+                    np.linspace(0, eps_lim, 10),
+                )
+            )
+
+        sig = self.get_stress(eps)
+        from structuralcodes.materials.constitutive_laws import UserDefined
+
+        # Return Marin coefficients for linearized version
+        return UserDefined(eps, sig).__marin__(**kwargs)
 
     def get_secant(self, eps: float) -> float:
         """Method to return the
--- a/structuralcodes/core/base.py
+++ b/structuralcodes/core/base.py
@@ -130,18 +130,21 @@ class ConstitutiveLaw(abc.ABC):
         sig[(sig < np.max(sig) * 1e-6)] = 0
         eps_lim = find_x_lim(eps, sig)
         # Now discretize the function in 10 steps for positive part
-        eps_pos = [] if eps_lim is None else np.linspace(0, eps_lim, 10)
+        eps_pos = (
+            np.linspace(0, -eps_min, 1)
+            if eps_lim is None
+            else np.linspace(0, eps_lim, 10)
+        )
         # Analise negative branch
         eps = np.linspace(0, eps_min, 10000)
         sig = -self.get_stress(eps)
         sig[(sig < np.max(sig) * 1e-6)] = 0
         eps_lim = find_x_lim(-eps, sig)
         # Now discretize the function in 10 steps for negative part
-        endpoint = bool(eps_pos == [])
         eps_neg = (
-            []
+            np.linspace(eps_min, 0, 1, endpoint=False)
             if eps_lim is None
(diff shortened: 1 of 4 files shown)


mortenengen:
I like the second option with the default `__marin__` which represents the `ConstitutiveLaw` as a piece-wise linear.

talledodiego:
> I like the second option with the default `__marin__` which represents the `ConstitutiveLaw` as a piece-wise linear.

Ok, I agree. Implemented this feature in [b220148](https://github.com/fib-international/structuralcodes/pull/88/commits/b2201485b06813262a9b1e507bd96c9aa5e831d8)

structuralcodes/materials/constitutive_laws.py
```diff
@@ -302,6 +302,87 @@ def get_ultimate_strain(
         return (100, self._eps_u)
 
 
+class Sargin(ConstitutiveLaw):
```
mortenengen, line 305: Just wondering, since it is meant to be used for concrete, should it be called `SarginConcrete`?

structuralcodes/materials/constitutive_laws.py
```diff
@@ -302,6 +302,87 @@ def get_ultimate_strain(
...
+    """Class for Sargin constitutive law.
+    The stresses and strains are assumed negative in compression
```
mortenengen, line 307: Please add a single line break before line 307.

structuralcodes/materials/constitutive_laws.py
```diff
@@ -302,6 +302,87 @@ def get_ultimate_strain(
...
+            are input, they will be assumed negative.
+        """
+        name = name if name is not None else 'ParabolaRectangleLaw'
+        super().__init__(name=name)
+        self._fc = -abs(fc)
+        self._eps_c1 = -abs(eps_c1)
+        self._eps_cu1 = -abs(eps_cu1)
+        self._k = k
+
+    def get_stress(self, eps: float) -> float:
```
mortenengen, line 345: The return type should be `ArrayLike`.

structuralcodes/materials/constitutive_laws.py
```diff
@@ -302,6 +302,87 @@ def get_ultimate_strain(
...
+
+        sig = self._fc * (self._k * eta - eta**2) / (1 + (self._k - 2) * eta)
+
+        # Elsewhere stress is 0.0
+        sig[eps < self._eps_cu1] = 0.0
+        sig[eps > 0] = 0.0
+
+        return sig
+
+    def get_tangent(self, eps: ArrayLike) -> ArrayLike:
```
mortenengen, line 359: There seems to be missing tests that cover this method.

structuralcodes/materials/constitutive_laws.py
```diff
@@ -302,6 +302,87 @@ def get_ultimate_strain(
...
+        tangent[eps > 0] = 0.0
+
+        return tangent
+
+    def get_ultimate_strain(
+        self, yielding: bool = False
+    ) -> t.Tuple[float, float]:
+        """Return the ultimate strain (positive and negative)."""
+        if yielding:
+            return (100, self._eps_c1)
```
mortenengen, line 382: Line 382 is not covered.

structuralcodes/materials/constitutive_laws.py
talledodiego: mmm.. I am not against it, but in general I find the names should be kept as the names of the model, otherwise we risk having SarginConcrete, KentParkConcrete, ThoeredolfConcrete etc...
Do you agree?

structuralcodes/core/base.py
```diff
@@ -102,9 +104,50 @@ def get_ultimate_strain(self) -> t.Tuple[float, float]:
...
+                return x[x_lim_index]
+            # All values are zero for x > 0
+            return None
+
+        eps_max, eps_min = self.get_ultimate_strain()
+        if eps_max > 1:
+            eps_max = 1
+        eps = np.linspace(eps_min, eps_max, 100000)
+        sig = self.get_stress(eps)
+        sig[(sig > 0) & (sig < np.max(sig) * 1e-6)] = 0
```
mortenengen, line 132: Please comment why L132 is necessary.

mortenengen (changes requested):
Very good 😃 I have left only a couple of comments that should be considered.

structuralcodes/materials/constitutive_laws.py
talledodiego: Done in [475c91c9](https://github.com/fib-international/structuralcodes/pull/88/commits/475c91c920213b94591c7b30431c873ad8bae3af) also for Parabola Rectangle and User Defined that had the same problem.

talledodiego: Fixed in [4ed7ae4a](https://github.com/fib-international/structuralcodes/pull/88/commits/4ed7ae4ab4dbc86da6fdeb49223f5dbc93136a47)

structuralcodes/materials/constitutive_laws.py
talledodiego: Fixed in [4ed7ae4a](https://github.com/fib-international/structuralcodes/pull/88/commits/4ed7ae4ab4dbc86da6fdeb49223f5dbc93136a47)

structuralcodes/materials/constitutive_laws.py
talledodiego: Fixed in [fe2bb3b](https://github.com/fib-international/structuralcodes/pull/88/commits/fe2bb3b0fbf26458420826a1edc07cdbe79a0926)

structuralcodes/materials/constitutive_laws.py
talledodiego: Done in [45984a9](https://github.com/fib-international/structuralcodes/pull/88/commits/45984a97f6990b94c1d8aea55e9d1fc51bd690cf)

structuralcodes/core/base.py
talledodiego: We need this line for the logic of the "smart" linearization. 
It may happen that we will have a positive branch in the constitutive law. For exponential for instance we have an exponential decay so the stress is never going to zero, but tend to it.
With L132 I set to 0 very small numbers of stress in the positive branch. This is done because the method find_x_lim finds the index where stress is back to zero again in order to discretize only the meaningful part of the stress-strain diagram.

mortenengen: Got it. Great if we could also add this check on the negative branch.

structuralcodes/materials/constitutive_laws.py
mortenengen: Let us leave it like it is.

talledodiego:
> Very good 😃 I have left only a couple of comments that should be considered.

Done, you can re-review it.

mortenengen (approved):
Very good!

PR #89, opened 2024-05-31
feat: First implementation of DB of Steel Sections

talledodiego:
First attempt of implementation of a database usable for Steel Sections. Implemented IPE, HE, IPN, UB, UC, UBP, IPN and UPN in this version. Data taken from Arcelormittal book with profile characteristics. Test performed with a relative tolerance of 2.5% to accomodate some differences that arise from section modulus reported in table and the more correct values obtained with our section analysis.


structuralcodes/geometry/_steel_sections.py
```diff
@@ -0,0 +1,1849 @@
...
+    get_geometry,
+    polygonize,
+    set_precision,
+)
+from shapely.affinity import scale, translate
+from shapely.geometry.polygon import orient
+from shapely.ops import linemerge, unary_union
+
+
+class IPE:
```
mortenengen, line 17: Just thinking, wouldn't it be a quite simple fix to do a Marin integration of the circumference of a section and obtain the section moduli? It would be great to see these as a set of properties 👍

mortenengen (changes requested):
This is a cool addition! I have left one question, and there are several lines that are not covered by testing. Please consider the question, and increase the coverage.

structuralcodes/geometry/_steel_sections.py
talledodiego: Ok, done an attempt in 7a9e5c16ca08ef181a46fe6c905af22891167297

talledodiego:
@mortenengen everything requested done:
1. added more test coverage in eb2fdb95859c886ef8cbcf0c5de04b1fae894982
2. added section moduli as properties in 7a9e5c16ca08ef181a46fe6c905af22891167297

Ready for your re-review

structuralcodes/geometry/_steel_sections.py
mortenengen: Good! Just a thought, instead of computing the properties in the constructor, could we do a check in the getter for the property, and compute the properties only if the internal attribute is `None`?

mortenengen:
> @mortenengen everything requested done:
> 
>     1. added more test coverage in [eb2fdb9](https://github.com/fib-international/structuralcodes/commit/eb2fdb95859c886ef8cbcf0c5de04b1fae894982)
> 
>     2. added section moduli as properties in [7a9e5c1](https://github.com/fib-international/structuralcodes/commit/7a9e5c16ca08ef181a46fe6c905af22891167297)
> 
> 
> Ready for your re-review

Very good! Three points before merging:

1. There was a conflict with the dev branch, I merged dev into this branch and pushed it, so please pull before commiting any fixes 😃 
2. I left a comment above regarding when to compute the section properties.
3. There are still a significant number of lines that are not covered by testing, please cover at least the _simple_ lines where we construct the names of the profiles.

structuralcodes/geometry/_steel_sections.py
talledodiego: You are absolutely right! I started doing like that because they started being very simple computations and I did not foresee a potential problem of performances.
Doing like you are proposing is for sure more efficient: you can find the consequent refactor in commit 6302c3949a7bb2e868b29da0b38c1ee560cb07df.

mortenengen:
Good job 😃 merged!

PR #103, opened 2024-06-10
feat: Calculate Gross Properties

talledodiego:
Draft implementation for computation of gross section properties.

### DIFF
--- a/structuralcodes/geometry/_geometry.py
+++ b/structuralcodes/geometry/_geometry.py
@@ -116,7 +116,12 @@ class PointGeometry(Geometry):
         Arguments:
             point: a couple of coordinates or e shapely Point object
             diameter: the diameter of the point
-            material: the material for the point
+            material: the material for the point (this can be a Material or
+                    a ConstitutiveLaw)
+            density (Optional: default None) when a ConstitutiveLaw is passed
+                    as material, the density can be providen by this argument.
+                    When the material is a Material object the density is taken
+                    from the material.
             name (Optional: default None): the name to be given to the object
             group_label (Optional: default None): a label for grouping several
         objects
@@ -238,7 +243,19 @@ class PointGeometry(Geometry):
         geo: PointGeometry,
         new_material: t.Optional[t.Union[Material, ConstitutiveLaw]] = None,
     ) -> PointGeometry:
-        """Create a new PointGeometry with a different material."""
+        """Create a new PointGeometry with a different material.
+
+        Arguments:
+        geo: (PointGeometry) the geometry
+        new_material: (Optional: dault None) a new material to
+            be applied to the geometry. If new_material is None
+            an Elastic material with same stiffness of the original
+            material is created.
+
+        Note:
+        The polygon is not copied, but just referenced in the
+        returned PointGeometry object.
+        """
         if not isinstance(geo, PointGeometry):
             raise TypeError('geo should be a PointGeometry')
         if new_material is not None:
@@ -315,7 +332,12 @@ class SurfaceGeometry:
 
         Args:
             poly (shapely.Polygon): Shapely polygon
-            mat (Material): A Material class applied to the geometry
+            mat (Material or ConstitutiveLaw): A Material or ConsitutiveLaw
+                class applied to the geometry
+            density (Optional: default None) when a ConstitutiveLaw is passed
+                    as mat, the density can be providen by this argument.
+                    When mat is a Material object the density is taken
+                    from the material.
         """
         # Check if inputs are of the correct type, otherwise return error
         if not isinstance(poly, Polygon):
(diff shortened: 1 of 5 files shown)


talledodiego:
@mortenengen I am asking your review here.

Points to discuss:
I computed area moments basing only on areas A, Sy, Sz, Iyy, Izz, Iyz (not accounting for different E of materials) and I have computed the same multiplied by their E (EA, ESy, ESz, EIyy, EIzz, EIyz). 
It's user's responsibility understanding if the former make sense to him.

We could also implement a `transform_gross_properties(E_ref)` method that return A, Sy, Sz, Iyy, Izz, Iyz etc. dividing each EA, ESy etc. by a reference value of E `E_ref` passed to the function. In this way we could obtain the homogenized values with elastic modulus of user choice (e.g. elastic modulus of concrete). 

Further topic: in computation of gross properties I have considered both SurfaceGeometry and PointGeometry (so I am accounting the effect of reinforcement bars). Is it ok or do we want just concrete (or better said just SurfaceGeometry)?
Finally, during computation of point geometry effects we are not cleaning the missing concrete. The error is generally small. Should we do it?

structuralcodes/geometry/_geometry.py
```diff
@@ -93,6 +104,7 @@ def __init__(
         point: t.Union[Point, ArrayLike],
         diameter: float,
         material: t.Union[Material, ConstitutiveLaw],
+        density: t.Optional[float] = None,
```
mortenengen, line 107: Should we perhaps specify that the value of density is only used if material is `ConstitutiveLaw`?

structuralcodes/geometry/_geometry.py
```diff
@@ -208,10 +228,43 @@ def rotate(
...
+            ):
+                raise TypeError(
+                    f'new_material should be a valid structuralcodes.base.\
+                    Material or structuralcodes.base.ConstitutiveLaw object. \
+                    {repr(new_material)}'
+                )
+        else:
+            # new_material not provided, assume elastic material with same
+            # elastic modulus
+            new_material = Elastic(E=geo.material.get_tangent(eps=0)[0])
```
mortenengen, line 257: This behaviour of creating a linear elastic material with the initial tangent as stiffness should be documented in the docstring.

structuralcodes/geometry/_geometry.py
```diff
@@ -208,10 +228,43 @@ def rotate(
...
             name=self._name,
             group_label=self._group_label,
         )
 
+    @staticmethod
+    def from_geometry(
+        geo: PointGeometry,
+        new_material: t.Optional[t.Union[Material, ConstitutiveLaw]] = None,
+    ) -> PointGeometry:
+        """Create a new PointGeometry with a different material."""
```
mortenengen, line 241: Please add a docstring.

structuralcodes/geometry/_geometry.py
```diff
@@ -607,3 +701,22 @@ def __sub__(self, other: Geometry) -> CompoundGeometry:
         for pg in self.point_geometries:
             processed_geoms.append(pg)
         return CompoundGeometry(geometries=processed_geoms)
+
+    @staticmethod
+    def from_geometry(
+        geo: CompoundGeometry,
+        new_material: t.Optional[t.Union[Material, ConstitutiveLaw]] = None,
+    ) -> CompoundGeometry:
+        """Create a new CompoundGeometry with a different material."""
```
mortenengen, line 710: Please describe the args and kwargs in the docstring.

structuralcodes/sections/_generic.py
```diff
@@ -96,9 +100,135 @@ def _calculate_gross_section_properties(self) -> s_res.GrossProperties:
...
+            warnings.warn(
+                'Perimiter computation for a multi polygon is not defined.'
+            )
+
+        gp.perimeter = polygon.exterior.length
+
+        # Computation of area: this is taken directly from shapely
         gp.area = self.section.geometry.area
+        # Computation of surface area, reinforcement area, EA (axial rigidity)
+        # and mass: Morten -> problem with units! how do we deal with it?
```
mortenengen, line 119: Good question. We specify in docstrings that the density should be given in kg/m3, but we are not specifying any unit for the geometry. I think it is ok for now, but we should do a more thorough review of all connected parts of the code before v0.0.1.

structuralcodes/sections/_generic.py
```diff
@@ -96,9 +100,135 @@ def _calculate_gross_section_properties(self) -> s_res.GrossProperties:
...
+        for geo in self.section.geometry.geometries:
+            gp.ea += geo.area * geo.material.get_tangent(eps=0)[0]
+            gp.area_surface += geo.area
+            if geo.density is not None:
+                # this assumes area in mm2 and density in kg/m3
+                gp.mass += geo.area * geo.density * 1e-8
```
mortenengen, line 125: L125 and L132, should it be 1e-9?

structuralcodes/sections/_generic.py
```diff
@@ -96,9 +100,135 @@ def _calculate_gross_section_properties(self) -> s_res.GrossProperties:
...
+
+        # Computation of E*S (first area moments * E)
+        # Implementation idea:
+        # Using integrator: we need to integrate integr(E*y*dA)
+        #    or integr(E*x*dA) for obtaining E*Sy and E*Sz.
+        #    Chat with Morten!
+        # Preparared small playground
+
+        def compute_area_moments(geometry, material=None):
```
mortenengen, line 141: This is smart 😃

mortenengen (changes requested):
Very good implementation 😃 just a couple of fixes that should be considered before merging.

talledodiego (commented):
Done everything you mentioned, thanks for the careful review!

mortenengen (approved):

PR #104, opened 2024-06-11
feat: First draft of MN Interaction Diagram

talledodiego:
- bugfix: Marin did not work correctly for constant strain distribution (curvature = 0)
- included first implementation for MN interaction domain 

TODO:

- [x]  implement NMM version
- [x] implement MM version
- [x] add corresponding tests

### DIFF
--- a/structuralcodes/core/_section_results.py
+++ b/structuralcodes/core/_section_results.py
@@ -110,22 +110,22 @@ class UltimateBendingMomentResults:
 
 
 @dataclass
-class NmmInteractionDomain:
+class NMMInteractionDomain:
     """Class for storing the NMM interaction domain results."""
 
-    n_theta: int = 0  # number of discretizations along the angle
-    n_axial: int = 0  # number of discretizations along axial load axis
+    num_theta: int = 0  # number of discretizations along the angle
+    num_axial: int = 0  # number of discretizations along axial load axis
 
     strains: ArrayLike = None  # array with shape (n,3) containing strains
     forces: ArrayLike = None  # array with shape(n,3) containing N, My, Mz
 
 
 @dataclass
-class NmInteractionDomain:
+class NMInteractionDomain:
     """Class for storing the NM interaction domain results."""
 
     theta: float = 0  # the inclination of n.a.
-    n_axial: float = 0  # number of discretizations along axial load axis
+    num_axial: float = 0  # number of discretizations along axial load axis
 
     n: ArrayLike = None  # Axial loads
     m_y: ArrayLike = None  # Moments My
@@ -133,10 +133,10 @@ class NmInteractionDomain:
 
 
 @dataclass
-class MmInteractionDomain:
+class MMInteractionDomain:
     """Class for storing the MM interaction domain results."""
 
-    n_theta: float = 0  # number of discretizations along the angle
+    num_theta: float = 0  # number of discretizations along the angle
     n: float = 0  # axial load
 
     theta: ArrayLike = None  # Angle theta respect axis Y
--- a/structuralcodes/core/_section_results.py
+++ b/structuralcodes/core/_section_results.py
@@ -116,6 +116,9 @@ class NmmInteractionDomain:
     n_theta: int = 0  # number of discretizations along the angle
     n_axial: int = 0  # number of discretizations along axial load axis
 
+    strains: ArrayLike = None  # array with shape (n,3) containing strains
+    forces: ArrayLike = None  # array with shape(n,3) containing N, My, Mz
+
 
 @dataclass
 class NmInteractionDomain:
@@ -135,3 +138,7 @@ class MmInteractionDomain:
 
     n_theta: float = 0  # number of discretizations along the angle
     n: float = 0  # axial load
+
+    theta: ArrayLike = None  # Angle theta respect axis Y
+    m_y: ArrayLike = None  # Moments My
+    m_z: ArrayLike = None  # Moments Mz
--- a/structuralcodes/core/_section_results.py
+++ b/structuralcodes/core/_section_results.py
@@ -107,3 +107,31 @@ class UltimateBendingMomentResults:
     chi_y: float = 0  # the curvature corresponding to the ultimate moment
     chi_z: float = 0  # the curvature corresponding to the ultimate moment
     eps_a: float = 0  # the axial strain at 0,0 corresponding to Mult
+
+
+@dataclass
+class NmmInteractionDomain:
+    """Class for storing the NMM interaction domain results."""
+
+    n_theta: int = 0  # number of discretizations along the angle
+    n_axial: int = 0  # number of discretizations along axial load axis
+
+
+@dataclass
+class NmInteractionDomain:
+    """Class for storing the NM interaction domain results."""
+
+    theta: float = 0  # the inclination of n.a.
+    n_axial: float = 0  # number of discretizations along axial load axis
+
+    n: ArrayLike = None  # Axial loads
+    m_y: ArrayLike = None  # Moments My
+    m_z: ArrayLike = None  # Moments Mz
+
+
+@dataclass
+class MmInteractionDomain:
+    """Class for storing the MM interaction domain results."""
+
+    n_theta: float = 0  # number of discretizations along the angle
+    n: float = 0  # axial load
--- a/structuralcodes/geometry/_geometry.py
+++ b/structuralcodes/geometry/_geometry.py
@@ -299,6 +299,18 @@ class SurfaceGeometry:
         """
         return self.polygon.centroid.coords[0]
 
+    def calculate_extents(self) -> tuple[float, float, float, float]:
+        """Calculate extents of SurfaceGeometry.
+
+        Calculates the minimum and maximum x and y values.
+
+        Returns:
+            Minimum and maximum x and y values (``x_min``, ``x_max``,
+            ``y_min``, ``y_max``)
+        """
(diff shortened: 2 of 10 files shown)


talledodiego:
with 3199c992cd926eedc43477386af6b8d4cfd6eb80 the full 3d NMM interaction domain is computed.

still todo:
1. tests implementation.
2. add a function for computing MM domain for given N.
3. Decide if we want to structure somehow the domain results.

talledodiego:
Included a new function for computing MM interaction given axial load `calculate_mm_interaction_domain`. 

Implemented a new function for MN interaction domain (`calculate_interaction_domain_2`) that is significantly faster (by almost a factor 10) respect to the first implementation. The reason of being so faster is because it avoids a lot of integrations.

structuralcodes/core/_section_results.py
```diff
@@ -107,3 +107,38 @@ class UltimateBendingMomentResults:
...
+@dataclass
+class MmInteractionDomain:
+    """Class for storing the MM interaction domain results."""
+
+    n_theta: float = 0  # number of discretizations along the angle
+    n: float = 0  # axial load
+
+    theta: ArrayLike = None  # Angle theta respect axis Y
+    m_y: ArrayLike = None  # Moments My
+    m_z: ArrayLike = None  # Moments Mz
```
mortenengen, line 144: Two small things to consider:
- Should the attrs `n_theta` and `n_axial` instead be called `num_theta` and `num_axial` to avoid confusion with the axial force `n`?
- Should the class names be `NMM...`, ``NM...`, and `MM...`?

structuralcodes/geometry/_geometry.py
```diff
@@ -527,6 +547,7 @@ def __init__(
                 for pg in self.point_geometries
             ]
             self.geom = MultiPolygon(geoms_representation)
+        self._reinforced_concrete = None
```
mortenengen, line 550: Remember to define this attr in the class header.

structuralcodes/sections/_generic.py
```diff
@@ -478,6 +581,233 @@ def calculate_moment_curvature(
 
         return res
 
+    def calculate_interaction_domain(
+        self, theta: float = 0, n_axial: int = 100
```
mortenengen, line 585: Same as for the result dataclass, could we say `num_axial` to avoid confusion with the axial force?

structuralcodes/sections/_generic.py
```diff
@@ -478,6 +581,233 @@ def calculate_moment_curvature(
...
+        eps_a = eps_n - kappa_y * y_n
+        kappa_z = np.zeros_like(kappa_y)
+
+        # rotate back components to work in section CRS
+        T = np.array([[cos(theta), -sin(theta)], [sin(theta), cos(theta)]])
+        components = np.vstack((kappa_y, kappa_z))
+        rotated_components = T @ components
+        return np.column_stack((eps_a, rotated_components.T))
+
+    def calculate_nmm_interaction_domain(
```
mortenengen, line 741: Same as for the result dataclass, could we say `num_axial` and `num_theta` to avoid confusion with the axial force?

structuralcodes/sections/_generic.py
```diff
@@ -478,6 +581,233 @@ def calculate_moment_curvature(
...
+            forces[i, 1] = My
+            forces[i, 2] = Mz
+
+        # Save to results
+        res.strains = strains
+        res.forces = forces
+
+        return res
+
+    def calculate_mm_interaction_domain(
```
mortenengen, line 782: Same as for the result dataclass, could we say `num_theta` to avoid confusion with the axial force?

structuralcodes/sections/_generic.py
```diff
@@ -478,6 +581,233 @@ def calculate_moment_curvature(
 
         return res
 
+    def calculate_interaction_domain(
```
mortenengen, line 584: To be consistent with the other methods, could this be called `calculate_nm_interaction_domain`?

structuralcodes/sections/_generic.py
```diff
@@ -478,6 +581,233 @@ def calculate_moment_curvature(
...
+        # integrate all strain profiles
+        forces = np.zeros_like(strains)
+        for i in range(strains.shape[0]):
+            N, My, Mz, tri = (
+                self.integrator.integrate_strain_response_on_geometry(
+                    geo=self.section.geometry,
+                    strain=strains[i, :],
+                    tri=self.triangulated_data,
+                )
+            )
```
mortenengen, line 646: Here, I guess we could do like?

```python
for strain in strains:
    N, My, Mz, tri = (
        self.integrator.integrate_strain_response_on_geometry(
            ...
            strain=strain,
            ...
        )
    )
```

structuralcodes/sections/_generic.py
```diff
@@ -478,6 +581,233 @@ def calculate_moment_curvature(
...
+        for i, n_i in enumerate(res.n):
+            res_bend_strength = self.calculate_bending_strength(
+                theta=theta, n=n_i
+            )
+            res.m_y[i] = res_bend_strength.m_y
+            res.m_z[i] = res_bend_strength.m_z
+
+        return res
+
+    def calculate_interaction_domain_2(
```
mortenengen, line 615: Please make this implementation our preferred one, and consider the method name and arg name above 😃

structuralcodes/sections/_generic.py
```diff
@@ -478,6 +581,233 @@ def calculate_moment_curvature(
...
+        # Initialize an empty array with the correct shape
+        strains = np.empty((0, 3))
+        for theta in thetas:
+            # Get ultimate strain profiles for theta angle
+            strain = self._compute_ultimate_strain_profiles(theta=theta)
+            strains = np.vstack((strains, strain))
+
+        # integrate all strain profiles
+        forces = np.zeros_like(strains)
+        for i in range(strains.shape[0]):
```
mortenengen, line 762: I guess we could do the same iteration as in the comment above.

mortenengen (changes requested):
Great contribution 😃 just some small comments for you to consider while finalizing.

structuralcodes/geometry/_geometry.py
talledodiego: Sorry, maybe I am misunderstanding the comment: what do you mean to define this in the class header? I think you are referring to add it before the __init__ right after the class definition. I am not understanding exactly why I should do that... Sorry for the for sure silly question..

structuralcodes/sections/_generic.py
talledodiego: I wasn't sure how the for .. in .. worked with a 2D array. I'v learnt something new! thanks!

structuralcodes/sections/_generic.py
talledodiego: We can delete the slow version if you want.. For now I am renaming the _2 as the default one and the old one as _old

structuralcodes/sections/_generic.py
talledodiego: As before...

talledodiego (commented):
Done in ebf8160186dc2b683d36f133e97bbb84f91f85e9

A doubt is still there about a for cycle that I am not sure I can do as you are suggesting. Can you check further please?

structuralcodes/geometry/_geometry.py
mortenengen: That is what I meant 😃 good to have it defined and typed one place, and with a reasonable default value.

structuralcodes/sections/_generic.py
mortenengen: Got it. You might do:

```python
for i, strain in enumerate(strains):
    N, My, Mz, tri = (
        self.integrator.integrate_strain_response_on_geometry(
            ...
            strain=strain,
            ...
        )
    )
```

structuralcodes/sections/_generic.py
mortenengen: Good 😃 yes, please remove the old one.

structuralcodes/sections/_generic.py
talledodiego: yes I know that, but it seemed not advantagueos..is it better for performance in you opinion?

mortenengen: I don't think the performance will be significantly improved, but I find the code slightly more readable 👍

talledodiego:
I should have done everything asked.

structuralcodes/sections/_generic.py
```diff
@@ -501,8 +498,9 @@ def calculate_limit_axial_load(self):
         y_n, y_p, strain = self.get_balanced_failure_strain(
             geom=self.section.geometry, yielding=False
         )
-        eps_p = strain[0] + strain[1] * y_p
-        eps_n = strain[0] + strain[1] * y_n
```
mortenengen: Please add a comment in the code why the small tolerance is included 👍

structuralcodes/sections/_generic.py
```diff
@@ -57,9 +57,16 @@ def __init__(
...
+        self._gross_properties = None
+
+    @property
+    def gross_properties(self) -> s_res.GrossProperties:
+        """Return the gross properties of the section."""
+        if self._gross_properties is None:
+            self._gross_properties = (
+                self.section_analyzer._calculate_gross_section_properties()
+            )
+        return self._gross_properties
```
talledodiego, line 69: Perfect! Thanks for having done that!

mortenengen (approved):

PR #109, opened 2024-06-25
feat: Implemented Popovics Constitutive Law

talledodiego:
Included Popovics-Mander  Constitutive Law and related test.

### DIFF
--- a/structuralcodes/materials/constitutive_laws.py
+++ b/structuralcodes/materials/constitutive_laws.py
@@ -130,13 +130,16 @@ class ElasticPlastic(ConstitutiveLaw):
     def get_stress(self, eps: ArrayLike) -> ArrayLike:
         """Return the stress given strain."""
         eps = np.atleast_1d(np.asarray(eps))
+        # Preprocess eps array in order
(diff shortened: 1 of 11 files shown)


structuralcodes/materials/constitutive_laws.py
```diff
@@ -461,6 +461,108 @@ def get_ultimate_strain(
         return (100, self._eps_cu1)
 
 
+class Popovics(ConstitutiveLaw):
+    """Class for Popovics-Mander constitutive law.
```
mortenengen, line 465: Please include a reference to the literature 👍

mortenengen (changes requested):
Very good 😃 Please add references to the literature for the Popovics (and also the Sargin) model, and we are good.

structuralcodes/materials/constitutive_laws.py
talledodiego: Done

talledodiego:
Added both references. Thanks for the review!

mortenengen (approved):

PR #110, opened 2024-07-02
feat: refine moment curvature

mortenengen:
A slight refinement of the `.calculate_moment_curvature` method on the `GenericSectionCalculator`.

- [x] Let the user specify the number of curvatures to calculate the response for.
- [x] Let the user specify the first curvature.
- [x] Let the user provide an array of curvatures.
- [x] Add tests.


talledodiego (approved):
Excellent contribution!
I like very much the usability of it!

PR #112, opened 2024-07-29
fix: typing of arg and return in get_stress and get_tangent

mortenengen:
Annotate both arg and return on `get_stress` and `get_tangent` of abstract `ConstitutiveLaw` as `ArrayLike` to be consistent with concrete implementations.


DanielGMorenaFhecor (approved):
Pretty straightforward change, but anyway, seems good for me! 💯

PR #125, opened 2024-08-05
feat: EC2 2023 8.6-Partially Loaded Areas

DanielGMorenaFhecor:
Small code for Chapter 8.6 from the new Eurocode 2:2023.


mortenengen (changes requested):
Thanks for the contribution!

I have merged dev and provided a minor fix to the docstring.

There are some lines that are not covered by tests. Please add tests so that we reach 100% coverage on the added lines.

structuralcodes/codes/ec2_2023/__init__.py
```diff
@@ -50,8 +50,10 @@
     wk_cal,
     wk_cal2,
 )
+from ._section_8_6_partially_loaded_areas import sigma_Rd_u
```
talledodiego, line 53: maybe it is better the name `sigma_Rdu`? The u is not a subscript of Rd

structuralcodes/codes/ec2_2023/_section_8_6_partially_loaded_areas.py
```diff
@@ -0,0 +1,65 @@
...
+    a1: float,
+    b: float,
+    ea: float = 0,
+    eb: float = 0,
+    nu_part: float = 3.0,
+) -> float:
+    """Calculate the design resistance for partially loaded areas without
+    horizontal force components.
+
+    EN1992-1-1:2023 Eq. (8.126).
```
talledodiego, line 19: Actually here there is not only (8.126) implemented, but also (8.127), (8.128), (8.129),...

A 1:1 implementation of (8.126) would require as input arguments in the function signature: fcd, Ac1, Ac0, nu_part

structuralcodes/codes/ec2_2023/_section_8_6_partially_loaded_areas.py
```diff
@@ -0,0 +1,65 @@
...
+        raise ValueError(f'b must not be negative. Got {b}')
+    if ea < 0:
+        raise ValueError(f'ea must not be negative. Got {ea}')
+    if eb < 0:
+        raise ValueError(f'eb must not be negative. Got {eb}')
+
+    # Eccentrically reduced loaded area
+    a0_red = a0 - 2 * ea
+    b0_red = b0 - 2 * eb
+    Ac0_red = a0_red * b0_red  # in mm²
```
talledodiego, line 58: Continuing above comment, for instance if we are in concentrically loaded area with don't need eccentiricity and determination of Ac0,red, so I find it strange having to specify this as a mandatory input.
One possible solution is refactor these equations into multiple functions. Alternatively we could use default values documenting it?
What do you think?

PR #113, opened 2024-07-29
structuralcodes - Repository examples

DanielGMorenaFhecor:
This draft pull request has been linked to the main repository to work on the examples 👍 . This pull request is not intended to be merged with the `dev` branch; it is solely for notifications and tracking purposes.

## Contributors
@MestreCarlos : Main contributor, working on examples proposed by @aperezcaldentey .
@DanielGMorenaFhecor : Reviewer, providing suggestions to improve the repository's usability from an end-user perspective.

## Initial Plan
We initially planned to create a separate repository for these examples and use `pip install git+[...]` to install the dev branch. However, this was not possible due to the missing `setup.py` file (correct me if I am wrong). Although adding the repo as a submodule was considered, we decided to continue working in the same repository. We will migrate to a different repository when the first version is released.

## Next Steps and Contributions

- Develop More Examples: We have created some examples, but additional examples are welcome. **These can be based on our experience, meetings with fib-TG10.3, or basic getting started examples**. A bullet list of current examples will be posted in the comments with the current developed ones—please feel free to suggest more.

- Improving Usability: @MestreCarlos and I have been discussing how to enhance the repository's usability from an end-user perspective. Notes from our discussions will be posted in the comments. Your feedback and contributions are welcome.

Feel free to comment, add, or discard any suggestions provided.

PR #132, opened 2024-08-07
Fix area computation on Generic Section

talledodiego:
This fix an issue mentioned in PR #113: the area should be the sum of surface + reinforcement area.

This code now returns the expected value:

```
# imports
from shapely import Polygon
from structuralcodes.materials.concrete import ConcreteEC2_2004
from structuralcodes.materials.reinforcement import ReinforcementEC2_2004
from structuralcodes.geometry import SurfaceGeometry
from structuralcodes.sections._reinforcement import add_reinforcement_line
from structuralcodes.sections import GenericSection

# Materials
concrete = ConcreteEC2_2004(25)
reinforcemnet = ReinforcementEC2_2004(fyk=500, Es=200000, density=7850, ftk=550, epsuk=0.07) 

# Create section
poly = Polygon(((0, 0), (350, 0), (350, 500), (0, 500)))
geo = SurfaceGeometry(poly, concrete)
geo = add_reinforcement_line(geo, (50, 450), (300, 450), 12, reinforcemnet, n=3)
geo = add_reinforcement_line(geo, (50, 50), (300, 50), 20, reinforcemnet, n=6)
sec = GenericSection(geo, integrator='marin')
# Print gross properties
print(f"area = {round(sec.gross_properties.area*1e-2)} cm2")
print(f"area_reinforcement = {round(sec.gross_properties.area_reinforcement*1e-2)} cm2")
print(f"area_surface = {round(sec.gross_properties.area_surface*1e-2)} cm2")
```

and the output is:
```
area = 1772 cm2
area_reinforcement = 22 cm2
area_surface = 1750 cm2
```

### DIFF
--- a/structuralcodes/sections/_generic.py
+++ b/structuralcodes/sections/_generic.py
@@ -142,6 +142,9 @@ class GenericSectionCalculator(SectionCalculator):
                 # this assumes area in mm2 and density in kg/m3
                 gp.mass += geo.area * geo.density * 1e-9
 
+        # Add area of reinforcement to total area
+        gp.area += gp.area_reinforcement
+
         # Computation of area moments
         #
         # Implementation idea:
--- a/tests/test_core/test_generic_section.py
+++ b/tests/test_core/test_generic_section.py
@@ -27,8 +27,16 @@ def test_rectangular_section():
     # The section
     poly = Polygon(((0, 0), (200, 0), (200, 400), (0, 400)))
     geo = SurfaceGeometry(poly, concrete)
-    geo = add_reinforcement_line(geo, (40, 40), (160, 40), 16, steel, n=4)
-    geo = add_reinforcement_line(geo, (40, 360), (160, 360), 16, steel, n=4)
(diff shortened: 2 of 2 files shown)


mortenengen:
I think I get it, but I am not sure if I find it intuitive.

For regular reinforcement embedded inside the concrete section, I would expect the area to be computed as the gross area of the concrete section. However, increased accuracy could be obtained by considering the concrete area as the gross area minus the reinforcement area, but this difference is usually neglected. We also rarely consider subtracting the reinforcement when calculating the area moments of the concrete. For external reinforcement, on the other hand, the total area should be the sum of the two.

Could you please elaborate a bit?

talledodiego:
You are right!
With the last commit c06b47ce336068ca6997a3fbec796557e66084f9 the API has been improved removing `area_surface` result from the gross properties. Now `area` contains the area of all surfaces while `area_reinforcement` contain area of all point geometries.

mortenengen (approved):
Good clarification.

PR #133, opened 2024-08-07
fix: bugfix marin coefficients

talledodiego:
Fixed a bug in parabola-rectangle constitutive law function for evaluation of Marin coefficients. The computation of the coefficients assumed the exponent of the parabolic part is equal to 2. If the exponent is different than 2, the marin method for all constitutive laws (even if less efficient) is called.

### DIFF
--- a/structuralcodes/materials/concrete/_concrete.py
+++ b/structuralcodes/materials/concrete/_concrete.py
@@ -4,10 +4,7 @@ import abc
 import typing as t
 
 from structuralcodes.core.base import ConstitutiveLaw, Material
-from structuralcodes.materials.constitutive_laws import (
-    ParabolaRectangle,
-    create_constitutive_law,
-)
+from structuralcodes.materials.constitutive_laws import create_constitutive_law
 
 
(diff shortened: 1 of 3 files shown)


mortenengen (changes requested):
Good fix 👍  could you please add a test that covers this line?

talledodiego:
> Good fix 👍 could you please add a test that covers this line?

Sure! Done in 374d9552831647aeae97d53ec932af78ee32a525

mortenengen:
Thanks for covering the line and also fixing the minor bugs. The bigger picture also looks cleaner now 😃 

About the warning. When do we encounter such a situation? Isn't this the parabolic branch of the parabolic rectangular law, which ends at `eps = eps_0`? Could it be that we are not encountering negative numbers, but something close to zero?

talledodiego:
Thanks!  think I got the small inconsistency! You pointed me to the right direction that I was stupidly not seeing!

mortenengen (approved):
Good job!

PR #159, opened 2024-09-13
Specialized sections

sshaave:
@talledodiego Hi. I have not had the time during the past months to look at this. I create a pull request so that we can keep track, as you suggested.

### DIFF
--- a/structuralcodes/sections/_rectangularRC.py
+++ b/structuralcodes/sections/_rectangularRC.py
@@ -29,9 +29,9 @@ def get_rebar_area(diameter: int) -> float:
     return diameter ** 2 / 4 * np.pi
 
 
-class RectangularCrossSection(GenericSection):
+"""class RectangularCrossSection(GenericSection):"""
 
-    """ A special case of GenericSection where the shape is rectangular, meant to both provide a more
+""" A special case of GenericSection where the shape is rectangular, meant to both provide a more
     user-friendly interaction with the code and run more optimized code where relevant.
 
     Attributes:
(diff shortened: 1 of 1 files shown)


structuralcodes/sections/_rectangularRC.py
```diff
@@ -0,0 +1,580 @@
...
+    ) -> None:
+        if name is None:
+            name = 'RectangularRCSection'
+        super().__init__(name)
+
+        # section geometry
+        self.height = height
+        self.width = width
+        self.cover = cover
+        self.geometry: SurfaceGeometry
```
sshaave, line 254: Here is where I start to change the class you started on. From what I see, there is not much to change or add

structuralcodes/sections/_rectangularRC.py
```diff
@@ -0,0 +1,580 @@
...
+    @property
+    def area(self) -> float:
+        """Returns Area of the cross section.
+        Only the gross area of concrete is computed.
+
+        Returns:
+            float: The area of concrete section.
+        """
+        return self.width * self.height
+
```
sshaave, line 311: I will add more methods, like get areal_reinforcement or get_area_concrete

structuralcodes/sections/_rectangularRC.py
```diff
@@ -0,0 +1,580 @@
...
+                # provided a Concrete material, used that one
+                self.confined_concrete = confined_concrete
+        else:
+            self.confined_concrete = concrete
+
+        # stirrups
+        if stirrups_data is None:
+            # If no stirrups are provided by default it is assumed the
+            # diameter is 8 mm - The spacing is not significant
+            stirrups_data = StirrupsData(8, 150)
```
sshaave, line 287: @talledodiego I agree that the diameter of the stirrups diameter is a needed parameter for the cross section, but the spacing seems misplaced in my opinion. I view the spacing of the reinforcement to be a parameter of the structure (i.e. beam or column). It could be either, it depends on the object hierarchy that is designed for this code.

PR #164, opened 2024-09-17
fix: Update geometry for consistent type hinting

talledodiego:


talledodiego:
@mortenengen I have done these small changes to better support type hinting related to geometry and to make consistent the API between `PointGeometry` and `SurfaceGeometry`.
It can be done even better probably (see comments along code).

structuralcodes/geometry/_geometry.py
```diff
@@ -712,7 +725,7 @@ def __init__(
             # useful for representation in svg
             geoms_representation = [g.polygon for g in self.geometries]
```
talledodiego, line 726: Here for instance polygon is not recognized by vscode. the reason is that `polygon` attribute is present for `SurfaceGeometry` while it is not available in `PointGeometry`. In a similar way `point` is available to `PointGeometry` and not to `SurfaceGeometry`.
Since `self.geometries` is a `list[Geometry]` type hinting is not working properly.
The alternatives I am seeing are:
1. We keep as it is as an acceptable implementation
2. See if it is possible to make it even more consistent the api. For instance defining a `geometry` attribute for the `Geometry` class that represents the point for a `PointGeometry` and a polygon for `SurfaceGeometry`.
I am afraid option 2 can be slightly complicated, but more importantly more confusing instead of having `point` and `polygon` attributes.
3. Delete `Geometry` class as a base class of `SurfaceGeometry` and `PointGeometry` and use `Union[SurfaceGeometry, PointGeometry` in the code instead of `Geometry`.

What do you think?

mortenengen: Good point. But if `geometries` can only by either a `MultiPolygon` or a `t.List[SurfaceGeometry]`, cannot we do:

4. `geometries: t.Union[t.List[SurfaceGeometry], MultiPolygon]`

?

If `geometries` can be a list of both points and surfaces, we should indicate this in the method signature and the docstring.

Also notice how `_process_geometries_list` is typed to return a tuple of two lists, and that this is the main reason for the polygon on line 726 is not recognized.

mortenengen (changes requested):
Good job 😃  Please consider the comment when finalizing the PR.

structuralcodes/geometry/_geometry.py
talledodiego: `geometries` can be a `MultiPolygon` or a list of any subclass of `Geometry` (i.e. `PointGeometry` and `SurfaceGeometry`).
For sure I need to update the docstring accordingly. I am not sure if I should change the signature or if it is correct already as:
```
def __init__(
        self,
        geometries: t.Union[t.List[Geometry], MultiPolygon],
        materials: t.Optional[t.Union[t.List[Material], Material]] = None,
    ) -> None:
```

Do you think the signature could work as it is?

mortenengen: I see, you are right, the signature is ok, but as you indicate, the docstring still says that `geometries` is either a list of `SurfaceGeometry` or a shapely `MultiPolygon` object, so this should be updated.

talledodiego: Sure, I corrected this in 3ec9527c7bc55527964535039d4cfcd17a70b7b4

talledodiego: Of course having `Geometry` in the signature my original question is still valid: is it ok if we loose type hinting in these lines? Since `Geometry` can be both `PointGeometry` or `SurfaceGeometry`

mortenengen (approved):

PR #176, opened 2024-09-24
docs: describe versioning

mortenengen:
Add a description of how we handle versioning in the docs. In short we say that the API of StructuralCodes is not yet stable. When the API is stable we will increase the major version to 1.0.0. Until then, we will indicate backwards incompatible changes by an increase of the minor version number.

This is loosly inspired by the current versioning scheme used by [Ruff](https://docs.astral.sh/ruff/versioning/).


talledodiego (approved):
Perfect! Thanks for having take care of this.

PR #180, opened 2024-09-26
bug: fix calculating Wely and Welz on steel sections

mortenengen:
Solves an issue where a call to `.Wely` or `.Welz` on a steel section raised a `TypeError` if not `.Iy` and `.Iz` was called from before.


talledodiego (approved):
Good catch!

PR #181, opened 2024-10-02
fix: swap order of strain limits returned from get_ultimate_strain

mortenengen:
The `get_ultimate_strain` method on `ConstitutiveLaw`s returns a tuple of strains. In the current implementation, it returns the positive strain first, and then the negative strain. It is perhaps more intuitive if we swap these, so that the returned strain limits are from left to right in a typical stress-strain diagram, or from low to high with respect to value. This is an attempt to swap the values.


talledodiego (approved):
Perfect. You are right it makes much more sense. I am not sure why I did it like that in the first place 🤣

PR #183, opened 2024-10-04
fix: moved methods for adding reinforcement into geometry module

talledodiego:
1) Moved methods for adding reinforcement inside the geometry module
2) Updated documentation accordingly


mortenengen (approved):
Thanks! This makes perfect sense 👍

PR #186, opened 2024-10-04
feat: more robust set design code

mortenengen:
Allows to pass a module to the `set_design_code` function, and improves the robustness of the method.


talledodiego (approved):
Excellent improvement on the robustness of the setting design code. I also liked the idea to pass a module!

PR #189, opened 2024-10-10
feat: add kwarg group_label to add_reinforcement functions

mortenengen:
Add the keyword argument `group_label` to the functions for adding reinforcement. This way, the created reinforcement is assigned the group label, and we prepare for other functions that can select based on the label.


talledodiego (approved):
Thanks for the addition, perfect!

PR #191, opened 2024-10-11
fix: use also relative tolerance when checking calculated Iyz

mortenengen:
In our calculation of the cross moment Iyz for generic sections, we raise a `RuntimeError` if the two ways of calculating the moment does not result in values that are close enough. In the current version we check only the absolute value of these results, but this is too strict if the section has a value of the cross moment significantly different from zero.

This is an attempt to liberalize the check by introducing a relative check in addition by using the `math.isclose` function. The absolute tolerance used in `isclose` is rather liberal, but the value is kept equal to the existing implementation.

Although beyond the scope of the PR, I have also changed `math.sin` and `math.cos` to the `numpy` versions of the functions to have a cleaner import block.

### DIFF
--- a/structuralcodes/sections/_generic.py
+++ b/structuralcodes/sections/_generic.py
@@ -5,7 +5,6 @@ from __future__ import annotations  # To have clean hints of ArrayLike in docs
 import math
 import typing as t
 import warnings
-from math import cos, sin
 
 import numpy as np
 from numpy.typing import ArrayLike
@@ -577,7 +576,12 @@ class GenericSectionCalculator(SectionCalculator):
         """Rotate triangulated data of angle theta."""
         rotated_triangulated_data = []
         for tr in self.triangulated_data:
-            T = np.array([[cos(theta), -sin(theta)], [sin(theta), cos(theta)]])
+            T = np.array(
+                [
+                    [np.cos(theta), -np.sin(theta)],
+                    [np.sin(theta), np.cos(theta)],
+                ]
+            )
             coords = np.vstack((tr[0], tr[1]))
             coords_r = T @ coords
             rotated_triangulated_data.append(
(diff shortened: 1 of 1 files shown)


structuralcodes/sections/_generic.py
```diff
@@ -208,7 +208,9 @@ def compute_area_moments(geometry, material=None):
             )
             # Change sign due to moment sign convention
             izz *= -1
-            if abs(abs(izy) - abs(iyz)) > 10:
+
+            # Check calculated cross moment
+            if not math.isclose(iyz, izy, rel_tol=1e-9, abs_tol=10):
```
talledodiego, line 213: ah! you can specify both relative and absolute tolerance then?

I didn't use relative because when they are small numbers it happens that relative tolerance can be high even if for engineering point of view they are basically the same.

This happens for instance whatever units you use (mm or m) when izy and iyz should be near zero for symmetric sections, probably.

I set the absolute to 10 thinking about working in mm (so inertia in mm4). But if you input m, I think 10 is too much? So we should set a value for absolute tolerance proportional to the size of the section.

For instance:
if everything is in mm, your section could be 400 x 600 mm. The inertia would have ordered of magnitude around 1e10 so 10 is about nine orders of magnitude lower. 
if everything is in m, your same section would be 0.4 x 0.6 m so the value 10 is too much: the inertia would be in the order of magnitude of 1e-2 so the absolute value should be around 1e-9? Am I correct?

structuralcodes/sections/_generic.py
```diff
@@ -574,7 +576,12 @@ def _rotate_triangulated_data(self, theta: float):
         """Rotate triangulated data of angle theta."""
         rotated_triangulated_data = []
         for tr in self.triangulated_data:
-            T = np.array([[cos(theta), -sin(theta)], [sin(theta), cos(theta)]])
+            T = np.array(
+                [
+                    [np.cos(theta), -np.sin(theta)],
```
talledodiego, line 581: I totally agree

talledodiego (changes requested):
Excellent. I only have a doubt for the value of absolute tolerance (see comment).

structuralcodes/sections/_generic.py
mortenengen: Good point. I have attempted an improvement in 133dbef.

talledodiego (approved):

PR #192, opened 2024-10-13
bug: fix ec2 2004 exponent in calculation of beta_ds

mortenengen:


talledodiego (approved):
Thanks for finding and correcting the bug.
On a more general comment (I already mentioned this in the original review of MC2010 code) I would split the function eps_cs (eq. 3.8) factoring out equations 3.9 - 3.10 - 3.13 - 3.12 - 3.11. I would consider this factoring out in the spirit of having a code equation covered by each function to permit better experimenting/research (for instance if the user wants to test an alternative definition of beta_as for instance he/she can use its own function and maintain the rest functions, while in the current approach he/she needs to rewrite the whole eps_cs function).

We could consider eventually this aspect in a separate PR.

I think the same correction done in #193 should be done here? (i.e. set t-ts = 0 when t is lower than ts)

mortenengen:
> Thanks for finding and correcting the bug. On a more general comment (I already mentioned this in the original review of MC2010 code) I would split the function eps_cs (eq. 3.8) factoring out equations 3.9 - 3.10 - 3.13 - 3.12 - 3.11. I would consider this factoring out in the spirit of having a code equation covered by each function to permit better experimenting/research (for instance if the user wants to test an alternative definition of beta_as for instance he/she can use its own function and maintain the rest functions, while in the current approach he/she needs to rewrite the whole eps_cs function).
> 
> We could consider eventually this aspect in a separate PR.
> 
> I think the same correction done in #193 should be done here? (i.e. set t-ts = 0 when t is lower than ts).

Thanks! For the further changes, I am on it in #194 😃

PR #193, opened 2024-10-14
fix: mc2010 beta_ds correct for time less than ts

mortenengen:
Correct the calculated value of `beta_ds` when the requested time is less than the time at start of exposure. This is necessary for correct computation of shrinkage development when an array of time from zero is provided.

### DIFF
--- a/structuralcodes/codes/mc2010/_concrete_creep_and_shrinkage.py
+++ b/structuralcodes/codes/mc2010/_concrete_creep_and_shrinkage.py
@@ -291,7 +291,9 @@ def beta_ds(
         numpy.ndarray: Multiplication factor used for calculating the drying
         shrinkage as a function of time.
     """
-    return np.sqrt((time - ts) / (0.035 * (notional_size) ** 2 + (time - ts)))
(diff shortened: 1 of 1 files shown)


talledodiego (approved):
Thanks for correcting this.

Again as a general comment, unrelated to your specific contribution in this PR, is that I would try to homogenize the implementation of MC2010 and EC2004. For instance beta_ds here is a function, in ec2004 it is integrated in eps_cs computation. Furthermore here it accepts an array of times that can be helpful for computing the factors for several times. I think we should try to aim some homogeneous implementation.

mortenengen:
> Thanks for correcting this.
> 
> Again as a general comment, unrelated to your specific contribution in this PR, is that I would try to homogenize the implementation of MC2010 and EC2004. For instance beta_ds here is a function, in ec2004 it is integrated in eps_cs computation. Furthermore here it accepts an array of times that can be helpful for computing the factors for several times. I think we should try to aim some homogeneous implementation.

Thanks! I am on top of the further changes in #194 😃

Issue #129, opened 2024-08-06
Homogenizing some code


talledodiego:
We should homogenize the code and specifically the public API to be consistent:

- [ ] accepting both positive and negative values when dealing with concrete quantities (e.g. eps_c1, fck, fcm, etc.) vs accepting only positive values and raising exception if negative.
- [ ] to be completed...

This homogenization should be done sooner than later in order to not change the public API after the release. 
It is true anyhow that version 0.0.1 has by definition an _unstable_ API that may be subjected to changes.

Maybe @mortenengen can get an idea of those inconsistencies now that he is working with docstrings for documentation? If this is the case you can add here your observations and we can start from here.

PR #196, opened 2024-10-16
feat: ec2-2004 expose crdc in shear resistance

mortenengen:
The parameter `CRdc` appears in Eq. (6.2) in EC2 (2004), but was not provided as an argument to the relevant function in `structuralcodes.codes.ec2_2004`.

Here, the parameter is added with a default value `None`. If no further value is provided, the default value from the design code `0.18 / gamma_c` will be used, otherwise, the provided value will be used.

### DIFF
--- a/structuralcodes/codes/ec2_2004/shear.py
+++ b/structuralcodes/codes/ec2_2004/shear.py
@@ -1,6 +1,7 @@
 """Design rules according to EN 1992-1-1 regarding shear."""
 
 import math
+import typing as t
 
 # General functions
 
@@ -171,6 +172,7 @@ def VRdc(
     fcd: float,
     k1: float = 0.15,
     gamma_c: float = 1.5,
+    CRdc: t.Optional[float] = None,
 ) -> float:
     """Compute the design strength of the shear resistance.
 
@@ -194,11 +196,13 @@ def VRdc(
             value might differ between National Annexes.
         gamma_c (float): Partial factor for concrete. Default value = 1.5,
             value might differ between National Annexes.
+        CRdc (Optional[float]): Scaling factor for the shear resistance.
+            Default value is 0.18 / gamma_c.
 
     Returns:
         float: The concrete shear resistance in MPa.
     """
-    CRdc = 0.18 / gamma_c
+    CRdc = CRdc or 0.18 / gamma_c
     return (
         max(
             CRdc * _k(d) * (100 * _rho_L(Asl, bw, d) * fck) ** (1.0 / 3.0)


talledodiego (approved):

talledodiego:
> The parameter `CRdc` appears in Eq. (6.2) in EC2 (2004), but was not provided as an argument to the relevant function in `structuralcodes.codes.ec2_2004`.
> 
> 
> 
> Here, the parameter is added with a default value `None`. If no further value is provided, the default value from the design code `0.18 / gamma_c` will be used, otherwise, the provided value will be used.

It is correct to expose the parameter. I am wondering, in analogy to what we thought in materials fcd for instance, how we could deal in these equations for the NA: CRdc, numin, k1 could have different expressions in NAs.

mortenengen:
> > The parameter `CRdc` appears in Eq. (6.2) in EC2 (2004), but was not provided as an argument to the relevant function in `structuralcodes.codes.ec2_2004`.
> > Here, the parameter is added with a default value `None`. If no further value is provided, the default value from the design code `0.18 / gamma_c` will be used, otherwise, the provided value will be used.
> 
> It is correct to expose the parameter. I am wondering, in analogy to what we thought in materials fcd for instance, how we could deal in these equations for the NA: CRdc, numin, k1 could have different expressions in NAs.
> 
> Also I am thinking if we should expose a function for vmin. (6.3N) is used here, but what if NA gives something different?

I totally see your points 👍

With respect to what functions to expose and not, we have a job to do to make all implementations homogeneous, and we are taking one step in e.g. #194 for creep and shrinkage from EC2 (2004). I updated and pinned #129.

The question about how to deal with NDPs, and especially those where the NA provides additional context than the main part, is something else that we should investigate more closely.

talledodiego:
> > > The parameter `CRdc` appears in Eq. (6.2) in EC2 (2004), but was not provided as an argument to the relevant function in `structuralcodes.codes.ec2_2004`.
> 
> > > Here, the parameter is added with a default value `None`. If no further value is provided, the default value from the design code `0.18 / gamma_c` will be used, otherwise, the provided value will be used.
> 
> > 
> 
> > It is correct to expose the parameter. I am wondering, in analogy to what we thought in materials fcd for instance, how we could deal in these equations for the NA: CRdc, numin, k1 could have different expressions in NAs.
> 
> > 
> 
> > Also I am thinking if we should expose a function for vmin. (6.3N) is used here, but what if NA gives something different?
> 
> 
> 
> I totally see your points 👍
> 
> 
> 
> With respect to what functions to expose and not, we have a job to do to make all implementations homogeneous, and we are taking one step in e.g. #194 for creep and shrinkage from EC2 (2004). I updated and pinned #129.
> 
> 
> 
> The question about how to deal with NDPs, and especially those where the NA provides additional context than the main part, is something else that we should investigate more closely.

I agree, it was a comment not strictly related to the PR per-se but to a more general reasoning we should do. And that potentially could break the API.

As a further comment that I forgot, Crd itself is a function with parameter gamma_c. So gamma_c is not directly involved in computation of 6.2, but on computation of Crd. In this PR you are taking as input Crd as a number (computed somewhere else) or by default it is 0.18/gamma_c. I would think about the API in this case. If a NA  uses a more complicated function for Crd depending not only on gamma_c it is somehow difficult. Maybe we could have as argument for the function Crd (as you have done) and not gamma_c. Then in another function compute default Crd as 0.018/gamma_c where default gamma_c is 1.5. If Crd is None then we call that function otherwise the user can use its own function.
But I think this is a matter of a future PR.

PR #200, opened 2024-10-22
feat: Rectangular Section

sshaave:
Pull request made for transparency, still more to be done I believe?


structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+    """
+
+    def __init__(self, height: float, width: float, concrete: Concrete, name: t.Optional[str] = None, integrator: t.Literal['marin', 'fiber'] = 'marin') -> None:
+        """Initialize a GenericSection.
+
+        Arguments:
+            geometry (Union(SurfaceGeometry, CompoundGeometry)): The geometry
+                of the section.
+            name (str): The name of the section.
+            integrator (str): The name of the SectionIntegrator to use.
```
talledodiego, line 37: should we include the default value selected?

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+    Attributes: TODO
+        geometry (Union(SurfaceGeometry, CompoundGeometry)): The geometry of
+            the section.
+        name (str): The name of the section.
+        section_calculator (GenericSectionCalculator): The object responsible
+            for performing different calculations on the section (e.g. bending
+            strength, moment curvature, etc.).
+    """
+
+    def __init__(self, height: float, width: float, concrete: Concrete, name: t.Optional[str] = None, integrator: t.Literal['marin', 'fiber'] = 'marin') -> None:
```
talledodiego, line 30: The material of a rectangular section is not concrete generally but any `Material`

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+from structuralcodes.materials.concrete import Concrete
+
+class RectangularSection(GenericSection): 
+    """This is the specific implementation of a rectangular class section.
+
+    The section is a 2D geometry where Y axis is horizontal while Z axis is
+    vertical.
+
+    The moments and curvatures around Y and Z axes are assumed positive
+    according to RHR.
```
talledodiego, line 19: We should also state that we create the rectangle centered on the origin?

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+    def __init__(self, height: float, width: float, concrete: Concrete, name: t.Optional[str] = None, integrator: t.Literal['marin', 'fiber'] = 'marin') -> None:
+        """Initialize a GenericSection.
+
+        Arguments:
+            geometry (Union(SurfaceGeometry, CompoundGeometry)): The geometry
+                of the section.
+            name (str): The name of the section.
+            integrator (str): The name of the SectionIntegrator to use.
+        """
+        poly: Polygon = Polygon(((0, 0), (width, 0), (width, height), (0, height)))
```
talledodiego, line 39: I think it is more efficient to create it already centered in (0,0) instead of later translating it?

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+        ea: float = area * geometry.material.get_tangent(eps=0)[0]
+        for geo in self.geometry.geometries:
+            pass
+        for geo in self.geometry.point_geometries:
+            pass
+
+        if geometry.density is not None:
+            mass = area * geometry.density * 1e-9
+        else:
+            mass = 0
```
talledodiego, line 62: Do we need to compute this stuff? Or is it already tackled by `GenericSection`?

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+
+
+        gp = s_res.GrossProperties()
+        gp.area = area
+        gp.perimeter = perimeter
+        gp.ea = ea
+        gp.mass = mass
+        # TODO: more gross sectional properties, and store them in self._gross_properties
```
talledodiego, line 70: I don't think we need to compute this stuff. It was here on the code I shared because we still did not developed `GenericSection`

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+
+    @property
+    def area(self) -> float:
+        """Returns Area of the cross section.
+        Only the gross area of concrete is computed.
+
+        Returns:
+            float: The area of concrete section.
+        """
+        return self.width * self.height
```
talledodiego, line 80: This can be removed for now. Here we have anyhow an example of something that in the future can be computed in a `RectangularSectionCalculator` for efficiency reasons. Instead of integrating with fibers or marin, we just multiply `height` and `width`.

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+
+
+class RectangularRC(Section):
```
talledodiego, line 83: what is `Section`? I think you mean `GenericSection`

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+    """Class for RC rectangular cross section.
+    The class assumes width and height respectively
+    along local x and y axes.
+
+    Input parameters:
+    width: width of the section
+    height: height of the section
+    cover: net cover (distance from outer edget of the section to
+    edge of the stirrup)
+    reinforcement_data: ReinforcementData
```
talledodiego, line 93: This seems not defined?

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+    concrete: Concrete
+    reinforcement: Reinforcement (TODO!!!!!! for now const law)
+    confined_concrete: Concrete (default = None) When set to None the
+    same concrete is adopted for concrete core. When using the string
+    "auto" the confinement will be automatically computed
+    stirrups_data: StirrupsData (default = None) to use when need
```
talledodiego, line 99: This dataclass seems not defined?

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+    """
+
+    def __init__(
+        self,
+        width: float,
+        height: float,
+        cover: float,
+        reinforcement_data: ReinforcementData,
+        concrete: Concrete,
+        reinforcement: ElasticPlastic,
```
talledodiego, line 138: this can be reinforcement.

talledodiego, line 138: I would mimic here the idea of `GenericSection`: we can input a `Material` or a `ConstitutiveLaw`. The `ConstitutiveLaw` is stored in the class.

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+                    confined_concrete.lower() == 'automatic'
+                    or confined_concrete.lower() == 'auto'
+                ):
+                    # TODO: compute automatic confinement
+                    if stirrups_data is None:
+                        raise ValueError(
+                            'Asked for automatic confinement \
+                            computation, but stirrups_data is None \
+                            \nPlease provide a proper stirrups_data'
+                        )
```
talledodiego, line 168: These lines and lines +179 to +182 are somehow incompatible: if `None` is provided we assume a default 182. We should behave coherently here and there.

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+            # diameter is 8 mm - The spacing is not significant
+            stirrups_data = StirrupsData(8, 150)
+        self.stirrups_data = stirrups_data
+
+        # longitudinal reinforcement
+        self.reinforcement_data = reinforcement_data
+        # arrange in the section the reinforcement
+        self._arrange_reinforcement_in_section()
+        # For now discretize in fibers here
+        self._discretize_fibers(10, 25)
```
talledodiego, line 190: I don't think we need for now to do discretization. 
We can add it later: we could avoid triangulation in the fiber integrator computing a quad triangulation.

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,446 @@
...
+        self.xs = np.array(self.xs)
+        self.ys = np.array(self.ys)
+        self.d = np.array(self.d)
+        # compute Areas for each reinforcement
+        self.As = np.pi * (self.d**2) / 4
+
+        self.xmin = xmin
+        self.xmax = xmax
+        self.ymin = ymin
+        self.ymax = ymax
```
talledodiego, line 446: I would remove these lines for now (see general comment)

tests/test_core/test_specific_section.py
```diff
@@ -0,0 +1,11 @@
...
+from structuralcodes.materials.concrete import ConcreteEC2_2004
+from structuralcodes.materials.reinforcement import ReinforcementEC2_2004
+
+
+# Test rectangular section
+def test_specific_rectangular_section():
+    """Test rectangular section."""
+    # Create materials to use
+    concrete = ConcreteEC2_2004(35)
+    steel = ReinforcementEC2_2004(fyk=500, Es=200000, ftk=450, epsuk=0.0675)
```
talledodiego, line 11: I think the test is incomplete?

talledodiego (changes requested):
Thanks @sshaave for your contribution on this topic that I find essential for our package.

As a first general comment, in every PR you should be sure that the pipeline is passing the checks. You can also run the tests in your local environment to solve potential problems. 
Then, when you find that the code can be reviewed, you can ask for review so that somebody can take a look as soon as possible. 

After these two general comments on PR contribution, I would like to ask you if we can simplify the code removing all the code with the algorithms (basically from line 194 in `specific.py`) and starting as a simple wrapper for `GenericSection`. Saying it differently, the computations will be carried out using our `GenericSectionCalculator`. In the future we will see if subclassing also `GenericSectionCalculator` to `RectangularRCSectionCalculator` if we see room for improvement of efficiency. But I suggest leaving this aspect for future PRs.

Then I am seeing (maybe I am not seeing correctly) that you are using as suggested the idea of having `ReinforcementData` and `StirrupsData` dataclasses for easy and flexible inputting of reinforcement parameters. But I am not seeing those classes in the code, probably you did not included those from the files that I sent you.

As a final comment, in general you should try to reach 100% test coverage on the files you are introducing.

Summarizing, you could try to focus on the init method of the derived classes (`RectangularSection` and `RectangularRCSection`) in this PR and the related dataclasses for inputing.

Are you willing to try this? If you need any assistance let me know how can I help you. Maybe we can set a meeting also with @mortenengen for better going through this?

structuralcodes/sections/specific.py
sshaave: I believe you are correct. How do we want to proceed? The idea behind the specific section is, as I understand it, to make it more efficient. Thus, calculating gross properties directly analytically would be the way to go?

structuralcodes/sections/specific.py
sshaave: Agree! I will change it

structuralcodes/sections/specific.py
sshaave: I forgot to communicate that my work is from the start and to line 83. Here, I have just pasted in your code.

sshaave:
@talledodiego Thanks for your comments. I forgot to communicate what I intended with this pull request, and thus I have created unneccessary work for you - sorry for that. What I have done is to define the class RectangularSection(GenericSection), and I pasted in your code from earlier so I could easier look at what you did. What I have done is basically define and write some calculations in calculate_gross_section_properties() for this class (which might be not necessary as of yet). What more do we want in this class? Maybe some dunder-classes?

structuralcodes/sections/specific.py
sshaave: Added following text for name name and integrator: Default value is set to None/to marin

structuralcodes/sections/specific.py
sshaave: Fixed

structuralcodes/sections/specific.py
sshaave: Agree and changed

structuralcodes/sections/specific.py
sshaave: Not relevant (this is your old code which I have pasted in as a help)

structuralcodes/sections/specific.py
sshaave: Not relevant (this is your old code which I have pasted in as a help)

structuralcodes/sections/specific.py
sshaave: Not relevant (this is your old code which I have pasted in as a help)

structuralcodes/sections/specific.py
sshaave: Not relevant (this is your old code which I have pasted in as a help)

structuralcodes/sections/specific.py
sshaave: Removed

tests/test_core/test_specific_section.py
sshaave: Correct

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,194 @@
...
+from structuralcodes.sections._generic import GenericSection
+from structuralcodes.geometry import SurfaceGeometry
+from structuralcodes.materials.concrete import Concrete
+from structuralcodes.core.base import Material
+
+class RectangularSection(GenericSection): 
+    """This is the specific implementation of a rectangular class section.
+
+    The section is a 2D geometry where Y axis is horizontal while Z axis is
+    vertical. The geometry has its center in origo
```
talledodiego, line 17:
```suggestion
    vertical. The geometry has its center in origin.
```

structuralcodes/sections/specific.py
talledodiego: I think in your change you created a square section instead of a rectangular section. You are referring only to `width` and not to `height`

structuralcodes/sections/specific.py
```diff
@@ -0,0 +1,194 @@
...
+    def __init__(self, height: float, width: float, material: Material, name: t.Optional[str] = None, integrator: t.Literal['marin', 'fiber'] = 'marin') -> None:
+        """Initialize a GenericSection.
+
+        Arguments:
+            geometry (Union(SurfaceGeometry, CompoundGeometry)): The geometry
+                of the section.
+            name (str): The name of the section. Default value is set to None
+            integrator (str): The name of the SectionIntegrator to use. Default value is set to marin
+        """
+        poly: Polygon = Polygon(((-width/2, -width/2), (width/2, -width/2), (width/2, width/2), (-width/2, width/2)))
```
talledodiego, line 40:
```suggestion
        poly: Polygon = Polygon(((-width/2, -height/2), (width/2, -height/2), (width/2, height/2), (-width/2, height/2)))
```

structuralcodes/sections/specific.py
talledodiego: I agree with your comment. I suggest to go one step at a time. As a first step let's just create a wrapper (no optimization intended). When this is working we will include the efficient computations.

In particular if we want to rewrite the `calculate_gross_section_properties` method we should compute all the values computed in the generic section. There is way more than just mass, area, ea.

Moreover I am not following here the code: why you compute area_reinforcement if this is a plain rectangular section?

structuralcodes/sections/specific.py
talledodiego: Same comment as before applies here.

structuralcodes/sections/specific.py
talledodiego: You did not change or remove the code here.
It could be fine for me to leave it like that as soon as it is coherent with the base class. In particular it seems to me that we don't have an area property in the current version of `GenericSection` but the area itself is inside the gross_properties that re computed with the section calculator....

I would remove this for now...

structuralcodes/sections/specific.py
talledodiego: Ok I get this, so you should remove it for now from this PR?

structuralcodes/sections/specific.py
talledodiego: Please try to keep the PR clean. You can have a local copy of the code I shared, but in the PR I suggest we leave only the code relevant for the PR itself and that eventually (at the end of the review process) will be merged with dev.

tests/test_core/test_specific_section.py
talledodiego: Ok, as soon as you go ahead with the code try to complete also the testing.

talledodiego (changes requested):
I think you still need to comply with all my comments so that we can resolve all of them.
I suggest in general to keep a clean version of the code in the PR so that eventually it could be merged.

In general as I first step, as already mentioned, I would just create a wrapper of the GenericSection to help the user in the user input phase. So in this phase I would not think about optimizing computations. We can leave this aspect for later, after this first contribution works.

It is essential than the pipeline passes all checks. You can check this in your local environment before committing or inspecting in github what checks did not pass. You should comply with those errors in order to make them pass. If you need any assistance in this step I am available for further guidance (and I am sure also @mortenengen is available).

I think we can always implement dunder methods or dunder classes. What do you have in mind here?

sshaave:
image

Any idea why the values n_max_marin and res_marin.n is both equal to 0.0?

structuralcodes/sections/specific.py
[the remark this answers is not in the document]
talledodiego: Yes, you can find an example of the code in `SurfaceGeometry`.

talledodiego:
> image
> 
> 
> 
> Any idea why the values n_max_marin and res_marin.n is both equal to 0.0?

I think it is because your rectangular section is a rectangle of homogeneous material. If you use concrete (that in this stage is imagined with zero tensile strength) there is no way it can take a tensile force. Therefore the value zero for both marin and fiber integrators.

talledodiego:
Hey there @sshaave !

During the last days me and @mortenengen worked on adding a further "specific section" that is a circular "section".
After some thinking we concluded that probably the wrapper class should not be a wrapper to a Section but a wrapper to a geometry. Saying it differently with these special methods we create geometries (or compound geometries) that then can be assigned to a GenericSection.

PR #207 does exactly that. Specifically I included a CircularGeometry that is a SurfaceGeometry derived class that permits to get a concrete surface to which apply some reinforcement. In the case of the circual geometry the pattern of reinforcement is easily inserted with the new function `add_reinforcement_circle`.

I propose to do the same in this PR for the rectangular case creating a `RectangularGeometry` that as you developed here takes as arguments b, h and the material.

It is better in our opinion to wrap around geometry and not section to better highlights the responsibilities here: with geometry we create the geometries, while we use the geometries in sections for doing the computations.

In the same PR #207 you can find also the relevant tests that I implemented for this new geometry.

What do you think? Are you willing to try modifying your code according to this suggestion?

Let me know if I can give you any further assistance on this.

sshaave:
Hello. I have looked through #207, and while I do not understand the practical differences between the approches, I am pleased to see implementations for other ubiquitous cross sections. I don't have the capacity to continue this work this year, and I have struggled to find time for structuralcodes lately. Feel free to rewrite the code for rectangular cross section/geometries.

talledodiego:
@sshaave @mortenengen I included in this PR the homogeneous RectangularSurface.
We will think in a future PR if there is the need for a RCRectangularGeometry. This PR can be considered completed.
@mortenengen ready for your review.

mortenengen (approved):
Good job on this one @sshaave and @talledodiego 😃 I will add the new class to the API docs before merging.

PR #202, opened 2024-10-22
fix: Constitutive law fixes and refactoring

talledodiego:
This PR covers two aspects:
1) Constitutive laws have been refactored as a sub-module permitting to have separate files for constitutive laws. @mortenengen maybe the file is already big enough to suggest a splitting into different files already?
2) when the user inputs a float in get_stress and get_tangent methods, a float will be returned. Otherwise a ArrayLike is returned.

The public API is not changed with this PR.


mortenengen (changes requested):
Thanks for the good and thorough fix!

I have only one thought: if `eps` or the return from `get_stress` or `get_tangent` is a `float`, is it still an `ArrayLike`? If not, the typing needs to be just slightly updated.

To your other question, yes, `_constitutive_laws.py` is already of significant size, so I suggest splitting it in one file per constitutive law.

talledodiego:
Thanks for the review. You are totally right!
Tried to fix in e6ff596bfb1ea350d98de1c084cda616cfc9a87f

mortenengen (approved):

mortenengen:
Nice job on this 😃

PR #203, opened 2024-10-22
fix: reinforced concrete property

talledodiego:
When translating or rotating a SurfaceGeometry or CompoundGeometry we have to pass also the concrete argument to the init method.


mortenengen (approved):
Nice that you spotted this bug!

Issue #105, opened 2024-06-13
Make units consistent


mortenengen:
For the moment, the units in e.g. sections, constitutive laws and code equations, are not consistent. This means that the user could input e.g. MPa for stresses, and kN for forces, without any guarantee that an infered value will have meaningful unit. We should explore options for handling this.

There are several options, including:
1. Not at all handling units, and letting it be up to the user to provide input in consistent units, and make sure that whenever unit dependent methods are implemented, the required unit is specified in the docstring.
2. We could set a global set of units that the user can adjust, expect that the user inputs values that are consistent with this set of units, and whenever a conversion is needed, this is performed automatically based on the globally set units.
3. We could set a global set of units like above, and all default values of input with units are handled automatically, such that in all calculations we are sure that the units are consistent, and that the result will have a consistent unit.

At least options 2 and 3 could required that we adapt some of the implemented design code equations to be compatible.

mortenengen:
Hey @connorferster. Thanks for reaching out, I really appreciate your comment on this issue. If you don't mind, I will continue the discussion on #205 where we have started a draft implementation.

PR #205, opened 2024-11-08
feat: handle units

mortenengen:
## Description

An attempt to handle units. Note: this is still a very immature draft!

Two classes are added to `core`: `UnitSet` and `UnitManager`. `UnitSet` is a `dataclass` responsible for storing a set of units intended to be used by a `Material` class. `UnitManager` uses two `UnitSet`s to convert between them.

No tests are implemented, and only a small set of units are implemented. `UnitManager` is not yet used by any `Material`, but we do have a plan for implementation.

## Example

This example shows how two sets of units are selected and passed to the unit manager. Furthermore, it shows how to convert stress, lengeth and force with the unit manager.

```python
from structuralcodes.core._units import UnitSet, UnitManager

default_units = UnitSet(length='mm', force='N')
alternative_units = UnitSet(length='m', force='kN')

manager = UnitManager(
    default_units=default_units,
    alternative_units=alternative_units,
)

manager.convert_stress_to_default(35e3)
manager.convert_length_from_default(30)
manager.convert_force_from_default(4e3)
```

### DIFF
--- a/structuralcodes/core/_units.py
+++ b/structuralcodes/core/_units.py
@@ -1,0 +1,135 @@
+"""The unit manager."""
+
+import typing as t
+from dataclasses import dataclass
+
+from numpy.typing import ArrayLike
+
+# Type annotations for units
+_LENGTH_LITERAL = t.Literal['mm', 'm']
+_FORCE_LITERAL = t.Literal['N', 'kN', 'MN']
+
+# Unit conversion
+_METER = 1.0
+_NEWTON = 1.0
+
+_UNITS = {
+    'length': {
+        'm': _METER,
+        'mm': _METER * 1e-3,
+    },
+    'force': {
+        'N': _NEWTON,
+        'kN': _NEWTON * 1e3,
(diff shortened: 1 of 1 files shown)


mortenengen:
@connorferster, please let me elaborate on your comment to #105 here.

In our group we have been discussing how to handle units for a while. We do not want to create an advanced library for unit handling, since these already exist, like e.g. [pint](https://pypi.org/project/Pint/) or [forallpeople](https://pypi.org/project/forallpeople/) which of course are very well aware of 😃 However, we are investigating how we can make it easier for our users to perform calculations with our material, geometry and section classes using different units, without introducing too much overhead for users and maintainers.

As you mention in your comment, it could be a good idea to leave unit handling outside of structuralcodes, let a well-established library deal with it, and instead make structuralcodes compatible with that library. But from our initial investigations, we realized that a minimum of internal implementation would be to handle length and force units and derive stress units from these, and we could do this on our material class.

In my first attempt to implement this, I actually used your forallpeople library. However, realizing that

PR #207, opened 2024-11-22
feat: Add circular geometries

talledodiego:
This PR includes a wrapper utility for creating circular geometries.

There is a `CircularGeometry` that represents a `SurfaceGeometry` with a circular shape. It can be created with:
```
circle = CircularGeometry(300, material)
```
Where 300 is the size of the section (by default it is the diameter, unless the argument `is_radius` is set to `True`) and material is a valid `Material` or `ConstitutiveLaw` object.

The resulting geometry can be further worked editing it with additions, subtractions, transformations. When the geometry is set, that can be assigned to a `GenericSection` to perform the computations.

Then a utility method inside `_reinforcement` named `add_reinforcement_circle` has been included permitting by default the distribution of `n` given bars (or given the spacing `s`) within the whole circumference. Passing `start_angle` and `stop_angle` the bars can be distributed in a circular arch instead that on the whole circumference. 

Finally a derived class of `CompoundGeometry` named `CircularRCGeometry` has been developed permitting with a very simple syntax to add the bars around the whole perimeter.

```
geo_rc = CircularRCGeometry(
    size=300,
    surface_material=material,
    diameter=12,
    number=8,
    cover=40,
    reinforcement_material=material,
    n_points=50,
)
geo_rc
```
which produces the following:
image

The alternative would have been creating methods instead of classes that return a Surface or Compound Geometry.

TODO:
- [x] Add tests when agreeing on this design

### DIFF
--- a/structuralcodes/geometry/_circular.py
+++ b/structuralcodes/geometry/_circular.py
@@ -1,17 +1,9 @@
-"""Classes for circular sections.
+"""Classes for circular geometries.
 
-The class `Circular` represents a circular section with homogenous material.
-The class `CircularRC` represents a reinforced concrete circular section
-with `Concrete` material for the `SurfaceGeometry` and `Reinforcement` material
-for the `PointGeometries`.
-
-These classes are simply wrappers of `GenericSection` class and permits an easy
-input by the user, but all calculations are performed in the under-the-hood
-`GenericSection` class.
-
-In the future this class will contain utility methods for automatically define
-confined concrete according to some confinement models. For now a not
-implemented error is raised if this feature is used.
+The class `CircularGeometry` represents a circular SurfaceGeometry with
+homogenous material.
+This class is simply a wrapper of `SurfaceGeometry` class and permits an easy
+input by the user.
 """
 
 import typing as t
--- a/structuralcodes/geometry/_circular.py
+++ b/structuralcodes/geometry/_circular.py
@@ -7,6 +7,7 @@ input by the user.
 """
 
 import typing as t
+from itertools import chain
 
 import numpy as np
 from shapely import Polygon
@@ -14,6 +15,7 @@ from shapely import Polygon
 from structuralcodes.core.base import ConstitutiveLaw, Material
 
 from ._geometry import CompoundGeometry, SurfaceGeometry
+from ._reinforcement import add_reinforcement_circle
 
 
 def _create_circle(radius, npoints=20):
@@ -100,10 +102,10 @@ class CircularRCGeometry(CompoundGeometry):
         diameter: float,
         number: int,
         reinforcement_material: t.Union[Material, ConstitutiveLaw],
+        cover: float,
         is_radius: bool = False,
         n_points: int = 20,
         surface_density: t.Optional[float] = None,
-        reinforcement_density: t.Optional[float] = None,
         concrete: bool = False,
     ) -> None:
         """Initialize a CircularRCGeometry.
@@ -118,6 +120,8 @@ class CircularRCGeometry(CompoundGeometry):
             number (int): The number of bars to distribute.
             reinforcement_material (Union(Material, ConstitutiveLaw)): A
                 Material or ConsitutiveLaw class applied to the point geometry.
+            cover (float): The cover of concrete intended as the distance
+                between outer line and barycenter of longitudinal bars.
             is_radius (bool): Indicates if size is interpreted as radius
                     (default = False).
             n_points (int): The number of points used to discretize the
@@ -126,10 +130,6 @@ class CircularRCGeometry(CompoundGeometry):
                 as surface_material, the density can be provided by this
                 argument. When surface_material is a Material object the
                 density is taken from the material.
-            reinforcement_density (Optional(float)): When a ConstitutiveLaw is
-                passed as reinforcement_material, the density can be provided
-                by this argument. When reinforcement_material is a Material
-                object the density is taken from the material.
             concrete (bool): Flag to indicate if the surface geometry is
                 concrete.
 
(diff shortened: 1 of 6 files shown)


structuralcodes/geometry/_circular.py
```diff
@@ -0,0 +1,173 @@
...
+from ._geometry import CompoundGeometry, SurfaceGeometry
+from ._reinforcement import add_reinforcement_circle
+
+
+def _create_circle(radius, npoints=20):
+    """Create a circle with a given radius."""
+    phi = np.linspace(0, 2 * np.pi, npoints + 1)
+    x = radius * np.cos(phi)
+    y = radius * np.sin(phi)
+    points = [(this_x, this_y) for this_x, this_y in zip(x, y)]
```
mortenengen, line 26: I suggest that we replace this with:

```python
points = np.transpose(np.array([x, y]))
```

It is about six times faster 😃

structuralcodes/geometry/_circular.py
```diff
@@ -0,0 +1,173 @@
...
+        material: t.Union[Material, ConstitutiveLaw],
+        is_radius: bool = False,
+        n_points: int = 20,
+        density: t.Optional[float] = None,
+        concrete: bool = False,
+    ) -> None:
+        """Initialize a CircularGeometry.
+
+        Arguments:
+            size (float): The size of the geometry. By default this number
```
mortenengen, line 49: Can we do either radius or diameter and use either `radius` or `diameter` as argument? Saves an argument, and makes it explicit for the user 👍 Since we are already using diameter when adding reinforcement, maybe we can use that also here?

talledodiego: I am not sure I got your proposal. You are suggesting to accept only `diameter` for instance and avoid this mechanism for accepting diameter or radius? Totally fine for me. I can do it for sure!

structuralcodes/geometry/_circular.py
talledodiego: I did not test it! but wow! for sure I will do it!

structuralcodes/geometry/_reinforcement.py
```diff
@@ -125,3 +122,127 @@ def add_reinforcement_line(
             group_label=group_label,
         )
     return geo
+
+
+def add_reinforcement_circle(
```
mortenengen, line 127: This is a good addition. Can we make it even more consistent with the `add_reinforcement_line` function? In the latter function, we have `geo, coords_i, coords_j, ...` as the first arguments in the signature, i.e. after the geometry we have arguments targeting the line to put reinforcement along. The equivalent in `add_reinforcement_circle` could be to have `geo, center, radius, start_angle, stop_angle, ...`. What do you think? But you probably implemented this order to have keyword arguments after positional arguments...

structuralcodes/geometry/_reinforcement.py
```diff
@@ -125,3 +122,127 @@ def add_reinforcement_line(
...
+            'Stop angle should be larger than start angle and difference \
+            them should be at most 2pi.'
+        )
+    # Calculate length from start_angle to stop_angle
+    length = radius * (stop_angle - start_angle)
+
+    # If the whole circle, than deal with the case that would add an extra bar
+    whole = False
+    add_n = 0
+    if abs(length - 2 * np.pi * radius) < 1e-4:
```
mortenengen, line 194: Does it make sense to instead use `math.isclose` here?

talledodiego: I am not sure I am following this suggestion...

structuralcodes/geometry/_circular.py
```diff
@@ -0,0 +1,173 @@
...
+        cover: float,
+        is_radius: bool = False,
+        n_points: int = 20,
+        surface_density: t.Optional[float] = None,
+        concrete: bool = False,
+    ) -> None:
+        """Initialize a CircularRCGeometry.
+
+        Arguments:
+            size (float): The size of the geometry. By default this number
```
mortenengen, line 114: Same comment as for `CircularGeometry`, could we do either `radius` or `diameter`?

structuralcodes/geometry/_circular.py
mortenengen: Exactly that 😃 At least to me, having just diameter, and skipping the `is_radius`, makes i clearer what to insert.

structuralcodes/geometry/_reinforcement.py
mortenengen: Got it. I am fine with that 👍

structuralcodes/geometry/_circular.py
talledodiego: Done in c29ed265e72637199db8ae79fae837c65cc47eca

structuralcodes/geometry/_circular.py
talledodiego: Sure. `is_radius` was optional anyhow so only if you wanted to insert the radius you would have need to worry about setting it. Anyhow I agree that the api is simpler when selecting just one. And the user can always do its division. I implemented only `diameter` in f635cb3b6e4c9f23fffbc0f3f034c32667f4db51. Anyhow I maintained both radius and diameter properties if those could be of any use.

structuralcodes/geometry/_circular.py
talledodiego: As agrred, I am removing CircularRC.

structuralcodes/geometry/_reinforcement.py
talledodiego: Sure! done in 87064519bacb2ab2efbe9ef68fb8e8426df07e90

talledodiego:
Done everything commented and added relevant tests.
Ready for the review @mortenengen

mortenengen (changes requested):
Thanks for this solid addition to the API. Only one thing missing, could you please add the new stuff to the API docs?

PR #160, opened 2024-09-16
feat: Cracked properties

MestreCarlos:
```python
from shapely import Polygon
from shapely.geometry import Polygon

from structuralcodes import codes, materials
from structuralcodes.geometry import SurfaceGeometry
from structuralcodes.sections._generic import GenericSection
from structuralcodes.sections._reinforcement import add_reinforcement_line

# from structuralcodes.plots.section_plots import draw_section_response,draw_section,get_stress_point

### CREATE THE SECTION ########
codes.set_design_code(design_code='ec2_2004')
concrete = materials.concrete.create_concrete(fck=25)
reinforcemnet = materials.reinforcement.create_reinforcement(
    fyk=500,
    Es=200000,
    density=7850,
    ftk=500,
    epsuk=0.07,
)
poly = Polygon(((0, 0), (350, 0), (350, 500), (0, 500)))
geo = SurfaceGeometry(poly, concrete)
geo = add_reinforcement_line(
    geo, (50, 450), (300, 450), 12, reinforcemnet, n=3
)
geo = add_reinforcement_line(geo, (50, 50), (300, 50), 20, reinforcemnet, n=6)
sec = GenericSection(
    geo,
)
sec.geometry = sec.geometry.translate(-175, -250)

)
### CREATE THE SECTION ########

print(sec.cracked_properties)
# print(sec.cracked_properties.i_yy_2 / sec.gross_properties.i11)
```

PR #210, opened 2024-12-03
feat: Elastic cracked properties

MestreCarlos:
Related with #160,  In this new branch I have implemented a function to obtain the cracking properties of a reinforced concrete section assuming linear elastic materials so that the result is independent of the stress state.


MestreCarlos:
```
import math

from shapely import Polygon

from structuralcodes import set_design_code
from structuralcodes.geometry import SurfaceGeometry, add_reinforcement
from structuralcodes.materials.concrete import create_concrete
from structuralcodes.materials.reinforcement import create_reinforcement
from structuralcodes.sections import GenericSection
from structuralcodes.sections._rc_utils import (
    calculate_elastic_cracked_properties,
)


# Set the active design code
set_design_code('ec2_2004')

# Create a concrete and a reinforcement
concrete = create_concrete(fck=45)
reinforcement = create_reinforcement(fyk=500, Es=200000, ftk=550, epsuk=0.07)

# Create a rectangular geometry
width = 250
height = 500
polygon = Polygon(
    [
        (-width / 2, -height / 2),
        (width / 2, -height / 2),
        (width / 2, height / 2),
        (-width / 2, height / 2),
    ]
)

geometry = SurfaceGeometry(poly=polygon, material=concrete)

# Add reinforcement
diameter_reinf = 16
cover = 50

geometry = add_reinforcement(
    geometry,
    (
        -width / 2 + cover + diameter_reinf / 2,
        -height / 2 + cover + diameter_reinf / 2,
    ),
    diameter_reinf,
    reinforcement,
)  # The add_reinforcement function returns a CompoundGeometry
geometry = add_reinforcement(
    geometry,
    (
        width / 2 - cover - diameter_reinf / 2,
        -height / 2 + cover + diameter_reinf / 2,
    ),
    diameter_reinf,
    reinforcement,
)

# Create section
section = GenericSection(geometry)
section.geometry

crack_prop, crack_sec = calculate_elastic_cracked_properties(section, 0.1 * math.pi)
crack_sec.geometry
```
Cracked Concrete Section Properties:
Total area: 21094.85870305712
Reinforcement area: 402.1238596594935
Axial rigidity (E * A): 713270533.0236123
Mass per unit length: 0.0031566722983270242
Perimeter: 681.6244256840238
First moment of area (Sy): 4237788.469779671
Second moment of area (Sz): 423072.5211365963
E * Sy: 114008331314.10426
E * Sz: 12692175634.097889
Centroid y Coordinate: 17.794336155027626
Centroid z Coordinate: 159.83883538664298
Second moment (Iyy): 915556886.2231364
Second moment (Izz): 111674189.75110063
Product moment (Iyz): 70069510.06605844
Centroidal Second moment (Iyy_c): 110045194.0203745
Centroidal Second moment (Izz_c): 103424370.57922992
Centroidal Product moment (Iyz_c): -11820265.553934209
Principal Second Moment (I11): 119009858.81790286
Principal Second Moment (I22): 94459705.78170156
Principal axis angle (theta): 0.6488648891660339
E * Iyy: 29986768560316.977
E * Izz: 3657098473554.968
E * Iyz: 2102085301981.7534
E * Iyy_c: 11763809658696.014
...
E * I11: 11764455871561.023
E * I22: 3430603420918.1685
Principal axis angle (theta): 0.008805834301442425
image

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+    curv = -1e-5  # Any curvature should return the same mechanical properties.
+    # Find the equilibrium with fixed curvature
+    eps = section.section_calculator.find_equilibrium_fixed_curvature(
+        rotated_geometry, 0, curv, 0
+    )[0]
+    z_na = -eps / curv  # distance to neutral fibre
+
+    # Cutting concrete geometries and retaining the compressed block
+    cut_geom = CompoundGeometry(None, None)
```
mortenengen, line 61: Could we here do the following:

```python
cut_geom = None
```

And then furthermore on L67 do:

```python
if cut_geom is None:
    ...
```

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        cut_geom += reinf
+
+    # return geoemtry to original rotation
+    cracked_geom = cut_geom.rotate(-theta)
+
+    # Define the cracked section
+    cracked_sec = GenericSection(cracked_geom)
+    cracked_prop = cracked_sec.gross_properties
+
+    print(f'{cracked_prop:cracked}')
```
mortenengen, line 83: Good if we can remove the `print` statement so that the function can be called without side-effects.

mortenengen (changes requested):
Great job on this one, @MestreCarlos 👍 I have left only a couple of minor comments that I hope you can have a look at. In addition, there are two lines not covered by the test, could you please add or extend the test to reach full coverage?

structuralcodes/core/_section_results.py
```diff
@@ -106,9 +106,20 @@ def __format__(self, spec: str) -> str:
...
+        if 'cracked' in spec.lower():
+            type = 'Cracked'
+            spec = spec.replace(
+                'cracked', ''
+            ).strip()  # remove "cracked" from spec
+        else:
+            type = 'Gross'
+
+        output_string = f'{type} Concrete Section Properties:\n'
+
```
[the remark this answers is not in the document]
talledodiego: I agree. I uses format here to permit simple format permitting like:
`print(f'{properties:.3f}')`

That's the reason why I originally created two dataclasses for `GrossProperties` and `CrackedProperties`. Being those very simple dataclasses it seemed a little bit too much to create a parent and derived classes as suggested by @mortenengen. Anyhow I am favorable to @mortenengen proposal if we feel more significant.

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
+from structuralcodes.geometry import CompoundGeometry, SurfaceGeometry
+from structuralcodes.materials.constitutive_laws import Elastic, UserDefined
+from structuralcodes.sections import GenericSection
+
+
+def calculate_elastic_cracked_properties(section: GenericSection, theta=0):
```
talledodiego, line 6: Probably should we type hint the output?

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+            surface_geometry = SurfaceGeometry(polygon, material)
+            # Add the new SurfaceGeometry to the list
+            surface_geometries.append(surface_geometry)
+
+        return surface_geometries
+
+    if not section.geometry.reinforced_concrete:
+        return None
+
+    rotated_geometry = section.geometry.rotate(theta)
```
talledodiego, line 41: I think we should rotate by -theta? see e.g. calculate bending strength method in the generic section.

The reason is because if we want to assess the inclination of n.a. equal to `theta` , to have n.a. horizontal, we should rotate the section by `-theta`:

image

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        Es = pg.material.get_tangent(eps=0)
+        elastic_steel = Elastic(Es, 'elastic steel')
+        pg._material = elastic_steel
+
+    curv = -1e-5  # Any curvature should return the same mechanical properties.
+    # Find the equilibrium with fixed curvature
+    eps = section.section_calculator.find_equilibrium_fixed_curvature(
+        rotated_geometry, 0, curv, 0
+    )[0]
+    z_na = -eps / curv  # distance to neutral fibre
```
talledodiego, line 58: I am not sure about this sign. The equation should be:

strain = strain0 + ky * z - kz * y

so n.a. should be:

z = strain / curv

?

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+            cut_geom = CompoundGeometry(subpart_sg)
+        else:
+            cut_geom += subpart_sg
+
+    # Add reinforcement geometries
+    for reinf in rotated_geometry.point_geometries:
+        cut_geom += reinf
+
+    # return geoemtry to original rotation
+    cracked_geom = cut_geom.rotate(-theta)
```
talledodiego, line 77: now the opposite: should be `theta` probably?

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Define the cracked section
+    cracked_sec = GenericSection(cracked_geom)
```
talledodiego, line 80: in this way we are not retaining the user desired integration method I think...
1) we should take the integrator and the mesh size from the original section. Furthermore the mesh_size being a percentage of the area, should be scaled accordingly.
2) is it better to return the compound geometry or the section?

talledodiego (changes requested):
Great! thanks very much for the good implementation! Just some minor comments in addition to @mortenengen  ones.

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        surface_geometries = []
+
+        # Iterate over the list of polygons and create SurfaceGeometry for each
+        for polygon in polygons_list:
+            # Create a new SurfaceGeometry for the current polygon
+            surface_geometry = SurfaceGeometry(polygon, material)
+            # Add the new SurfaceGeometry to the list
+            surface_geometries.append(surface_geometry)
+
+        return surface_geometries
```
talledodiego, line 36: Why we return a list of Surface geometries here and not directly a `CompoundGeometry`?

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    eps = section.section_calculator.find_equilibrium_fixed_curvature(
+        rotated_geometry, 0, curv, 0
+    )[0]
+    z_na = -eps / curv  # distance to neutral fibre
+
+    # Cutting concrete geometries and retaining the compressed block
+    cut_geom = CompoundGeometry(None, None)
+    for i, part in enumerate(rotated_geometry.geometries):
+        min_x, max_x, min_y, max_y = part.calculate_extents()
+        upper_div, lower_div = part.split(((min_x, z_na), 0))
```
talledodiego, line 64: Maybe I am wrong, but if I remember well, you could avoid line 63 with calculating extent and use `split(((0, z_na), 0))`. I think that the method `split` will create the correct line taking into account the boundaries. Since it does anyhow that computation L63 could be redundant.

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        # Convert to SurfaceGeometry
+        subpart_sg = create_surface_geometries(upper_div, part.material)
```
talledodiego, line 66: if we return a `CompoundGeometry` here we could just sum to the old `CompundGeometry` (with the exception of `i == 0` for which we inizialize it). I think maybe you could avoid the enumerate and check just if cut_geom is None:

```
if cut_geom is None:
    cut_geom = subpart_sg
else:
    cut_geom += subpart_sg
```

talledodiego (changes requested):
I have included a couple of minor aspects also.

structuralcodes/sections/_rc_utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # return geoemtry to original rotation
+    cracked_geom = cut_geom.rotate(-theta)
+
+    # Define the cracked section
+    cracked_sec = GenericSection(cracked_geom)
+    cracked_prop = cracked_sec.gross_properties
+
+    print(f'{cracked_prop:cracked}')
+    return cracked_prop, cracked_sec
```
mortenengen, line 84: Can we do a slight change here? It would be good if we by default only return the `cracked_prop`. To control if we should return the `cracked_sec` in addition, we should have an additional keyword argument, e.g. `return_cracked_section`, which by default is `False`. So by default, the return object is `CrackedProperties`, and if `return_cracked_section is True`, the return object is `t.Tuple[CrackedProperties, GenericSection]`.

structuralcodes/core/_section_results.py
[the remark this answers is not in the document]
MestreCarlos: I agree. It is not an appropriate use of spec. My mistake for not being familiar with its standard use in Python.

Personally, I think the mechanical properties related to the theta angle do not have a practical use. If needed, the section can always be rotated first, and then the cracked properties can be requested with the argument theta=0.

That said, I think the simplest solution would be to remove the word "gross" in the __format__ function of GrossProperties and rename the class GrossProperties to SectionProperties. This way, we keep a single class and don't complicate this matter too much. 

Even so, I am open to whatever you prefer.

structuralcodes/sections/_rc_utils.py
MestreCarlos: Much better. Thanks.

structuralcodes/sections/_rc_utils.py
MestreCarlos: You are right.

structuralcodes/sections/_rc_utils.py
MestreCarlos: Better. This way, we maintain sign consistency within the package.

structuralcodes/sections/_rc_utils.py
MestreCarlos: Perhaps it would be better to return cracked_geom (CompoundGeometry) directly and not the section. I'm not sure if returning the complete GenericSection object adds any value. What do you think?

structuralcodes/sections/_rc_utils.py
MestreCarlos: You are right

structuralcodes/sections/_rc_utils.py
MestreCarlos: Ok. I will do it

structuralcodes/core/_section_results.py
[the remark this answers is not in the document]
mortenengen: I agree with you @MestreCarlos. Would you like to remove the "gross" and rename the class?

structuralcodes/sections/_rc_utils.py
mortenengen: Good. Please remember to update the hint after renaming the properties-class.

structuralcodes/sections/_rc_utils.py
mortenengen: This makes perfect sense, @MestreCarlos. As I see it, the reason for returning either the section or the geometry is for letting the user check what was the basis for the calculations, and the geometry is the exactly the basis for this.

However, when the `cracked_sec` is initialized, the integrator, and mesh size if any, should be inherited from the section that was input to the function in order to maintain user control, as commented by @talledodiego above.

mortenengen (approved):

mortenengen:
Good job on this one, @MestreCarlos! Thanks for the contribution 😃  I am awaiting approval from @talledodiego before merging with dev.

structuralcodes/core/_section_results.py
```diff
@@ -106,20 +106,9 @@ def __format__(self, spec: str) -> str:
...
-            type = 'Cracked'
-            spec = spec.replace(
-                'cracked', ''
-            ).strip()  # remove "cracked" from spec
-        else:
-            type = 'Gross'
-
-        output_string = f'{type} Concrete Section Properties:\n'
-
+        output_string = 'Concrete Section Properties:\n'
```
talledodiego, line 111: I missed this from my previous contributions, but should the title should be "Section Properties" probably?

structuralcodes/core/_section_results.py
```diff
@@ -131,7 +120,7 @@ def __format__(self, spec: str) -> str:
         return output_string
 
     def __str__(self) -> str:
-        """Returns the informal string representation of the gross concrete
+        """Returns the informal string representation of the concrete
```
talledodiego, line 123: same here, we should remove the concept of "concrete" now that we have a generic SectionProperty?

structuralcodes/sections/_rc_utils.py
```diff
@@ -58,27 +74,41 @@ def create_surface_geometries(polygons_list, material):
...
+        mesh_size = getattr(
+            section.section_calculator.integrator, 'mesh_size', 0.01
+        )
+        cracked_sec = GenericSection(
+            cracked_geom,
+            integrator='fiber',
+            mesh_size=mesh_size,
+        )
+    else:
+        cracked_sec = GenericSection(cracked_geom, integrator='marin')
```
talledodiego, line 105: not your fault here, you had to write it like this because the API of the section is a little bit lacking methods for doing this kind of operations so you had to do in this intricate way... Probably I will improve the API after merging this...

talledodiego (approved):
great work!

structuralcodes/core/_section_results.py
mortenengen: Fixed in c453e63.

structuralcodes/core/_section_results.py
mortenengen: Fixed in c453e63.

structuralcodes/sections/_rc_utils.py
mortenengen: Good point. I have created a new issue (#211) for this and assigned it to you @talledodiego.

PR #213, opened 2025-01-15
feat: add origin kwarg to wrapper geom classes

mortenengen:
I have added a kwarg to both the `CircularGeometry` and the `RectangularGeometry` to easily set the origin of the geometry to a point other than `(0.0, 0.0)`.

### DIFF
--- a/tests/test_geometry/test_circular.py
+++ b/tests/test_geometry/test_circular.py
@@ -40,6 +40,37 @@ def test_create_circular_geometry_exception(wrong_diameter):
         CircularGeometry(wrong_diameter, mat)
 
 
+# Test providing origin
+@pytest.mark.parametrize(
+    'origin',
+    [(0, 0), (30, 600), (1.0,)],
+)
+def test_circle_with_origin(origin):
+    """Test creating a circle with an origin."""
+    # Arrange
+    mat = Elastic(300000)
+    diameter = 100
+
+    # Act and assert
+    if origin is not None and len(origin) == 2:
+        geom = CircularGeometry(diameter=diameter, material=mat, origin=origin)
+        assert np.allclose(
+            (geom.polygon.centroid.xy[0][0], geom.polygon.centroid.xy[1][0]),
+            origin,
+        )
+    elif origin is None:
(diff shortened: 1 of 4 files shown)


tests/test_geometry/test_circular.py
```diff
@@ -40,6 +40,37 @@ def test_create_circular_geometry_exception(wrong_diameter):
...
+        assert np.allclose(
+            (geom.polygon.centroid.xy[0][0], geom.polygon.centroid.xy[1][0]),
+            origin,
+        )
+    elif origin is None:
+        geom = CircularGeometry(diameter=diameter, material=mat, origin=origin)
+        assert np.allclose(
+            (geom.polygon.centroid.xy[0][0], geom.polygon.centroid.xy[1][0]),
+            (0, 0),
+        )
```
talledodiego, line 66: With the current parameters in parametrize, does this occur? Should we add None in one of the parametrised tests?

tests/test_geometry/test_rectangular.py
```diff
@@ -51,6 +52,42 @@ def test_create_rectangular_geometry_exception(wrong_width, wrong_height):
...
+            origin,
+        )
+    elif origin is None:
+        geom = RectangularGeometry(
+            width=width, height=height, material=mat, origin=origin
+        )
+        assert np.allclose(
+            (geom.polygon.centroid.xy[0][0], geom.polygon.centroid.xy[1][0]),
+            (0, 0),
+        )
```
talledodiego, line 83: Same comment as before.

talledodiego (approved):
Excellent, thanks for having picked up this enhancement!

tests/test_geometry/test_circular.py
mortenengen: Yes, that was my idea, but it slipped. Thanks for noticing! Fixed in 23441f4.

tests/test_geometry/test_rectangular.py
mortenengen: Same as above 😄

talledodiego (approved):

Issue #214, opened 2025-01-21
Missing kwargs and call to super().__init__ in geometry creation


mortenengen:
When initializing `SurfaceGeometry`, `RectangularGeometry`, and `CircularGeometry` we are missing the kwargs `name` and `group_label`.

And in `SurfaceGeometry` we are missing the call to `super().__init__()`.

@talledodiego, do you agree that this should have been included?

talledodiego:
You are totally right. I think I simply forgot about it. Excellent if you can add this!

PR #217, opened 2025-01-23
fix: add missing kwargs in geometries

mortenengen:
Solves #214.

### DIFF
--- a/structuralcodes/geometry/_circular.py
+++ b/structuralcodes/geometry/_circular.py
@@ -42,6 +42,8 @@ class CircularGeometry(SurfaceGeometry):
         density: t.Optional[float] = None,
         concrete: bool = False,
         origin: t.Optional[ArrayLike] = None,
+        name: t.Optional[str] = None,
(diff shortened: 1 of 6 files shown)


talledodiego (commented):
Thanks @mortenengen ! I have just one thing in mind that I would like to consider one second before closing this:
Should we (like in other cases for instance for material names) propose some defaults for names? (in particular when using for instance circular or rectangular... What do you think?

mortenengen:
> Thanks @mortenengen ! I have just one thing in mind that I would like to consider one second before closing this:
> Should we (like in other cases for instance for material names) propose some defaults for names? (in particular when using for instance circular or rectangular... What do you think?

Good point @talledodiego. I was considering it, but since the base geometry class assigns a default name, I did not follow it further. Do you think we should be more specific?

talledodiego:
No no, I don't have strong reasons for being more specific , just came in my mind. 
I suggest merging this as is, maybe we can rethink this in the future or f we need to be more specific.

PR #224, opened 2025-03-06
fix: bugfix moment curvature

talledodiego:
In some specific cases the moment curvature iterative algorithm fails.

This is due to the prefind range where to apply bisection algoirhtm.

A more robust implementation is proposed.

### DIFF
--- a/structuralcodes/sections/_generic.py
+++ b/structuralcodes/sections/_generic.py
@@ -443,8 +443,12 @@ class GenericSectionCalculator(SectionCalculator):
         found = False
         it = 0
         delta = 1e-3
+        # Use a growth factor for an exponential finding
+        r = 2.0
+        diverging = False
+        diverging_steps = 0
         while not found and it < ITMAX:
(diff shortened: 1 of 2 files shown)


structuralcodes/sections/_generic.py
```diff
@@ -457,10 +461,19 @@ def _prefind_range_curvature_equilibrium(
...
                 # to failure of a material
-                delta /= 2
-                it -= 1
+                diverging = True
+                if diverging:
+                    # Count for how many steps we are diverging
+                    diverging_steps += 1
+                # If we are consistently diverging for more than 10 steps,
+                # Restart the process with a small delta
+                if diverging_steps > 10:
```
mortenengen, line 472: This is a nice fix. Could we strengthen it by adding a maximum number of restarts just to avoid an infinite `while`?

mortenengen (changes requested):
Thanks for providing this fix! I have left one small comment for you to consider.

mortenengen (approved):

PR #226, opened 2025-03-11
feat: make materials immutable

mortenengen:
## Description of actions taken

- Removed the `_reset_attributes` and the `_update_attributes`methods.
- Initialized the constitutive law in `__init__`. Skipped the check for material in the constitutive law, but the available laws out-of-the-box are reflected by the literal type annotation in the signature. The implementation allows passing a string representation of one of the constitutive laws that we have defined dunder-methods for, but also opens for passing directly a custom constitutive law. I.e. full flexibility.
- Rewrote `create_constitutive_law` to be more robust with respect to catching errors and to avoid returning `None`.
- All attributes are initialized either to `None` or the provided value inside `__init__`.
- All attribute validation is performed in a separate `__post_init__` which is called at the end of `__init__`. All validation that was previously performed in the setters is moved to this method.
- Removed check of valid materials for the provided constitutive law. The reason is that the constitutive law requires valid attributes in order to be correctly initialized, so the attribute validation is performed before the constitutive law is initialized.
- All properties are rewritten to the form *if None return code value else return provided value*. No caching, but clear what attributes are overridden in `__init__`, and most important, all materials have consistent behavior, and the non-propagating-override-bug is eliminated.
- Added notes in docstrings of all properties stating what other properties they are derived from if they are not provided when initializing the object.
- Removed all setters.
- Homogenized the code of the reinforcement classes such that these materials behave consistent with the concretes. This is mainly related to how the constitutive law is initialized.
- Fixed bug in the default constitutive law of reinforcement. Before, the hardening modulus was based on the characteristic stress-strain diagram, after, it is based on the design diagram which is consistent with the implementation in `__elasticplastic__`.


structuralcodes/materials/concrete/_concrete.py
```diff
@@ -34,72 +33,17 @@ def __init__(
...
-        """Each concrete should define its own _reset_attributes method. This
-        is because fck setting, reset the object arguments.
-        """
-
     @property
     def constitutive_law(self) -> ConstitutiveLaw:
         """Returns the constitutive law object."""
-        if self._constitutive_law is None:
-            self.constitutive_law = 'parabolarectangle'
         return self._constitutive_law
```
talledodiego, line 45: Is this now necessary since it is the same implementation of `Material` class?

structuralcodes/materials/concrete/_concreteEC2_2004.py
```diff
@@ -34,6 +34,31 @@ def __init__(
...
+        fctk_95: t.Optional[float] = None,
+        Ecm: t.Optional[float] = None,
+        eps_c1: t.Optional[float] = None,
+        eps_cu1: t.Optional[float] = None,
+        k_sargin: t.Optional[float] = None,
+        eps_c2: t.Optional[float] = None,
+        eps_cu2: t.Optional[float] = None,
+        n_parabolic_rectangular: t.Optional[float] = None,
+        eps_c3: t.Optional[float] = None,
+        eps_cu3: t.Optional[float] = None,
```
talledodiego, line 61: Now this list is quite limited. I think we expect this could grow up... (that really was the reasoning behind all these setters making it mutable)...

Is there a smart way to avoid that init attributes explode?

structuralcodes/materials/concrete/_concreteEC2_2004.py
```diff
@@ -62,140 +132,197 @@ def __init__(
...
+            if n_parabolic_rectangular is not None
+            else None
+        )
+        self._eps_c3 = abs(eps_c3) if eps_c3 is not None else None
+        self._eps_cu3 = abs(eps_cu3) if eps_cu3 is not None else None
+
+        self.__post_init__()
+
+        # The constitutive law requires valid attributes, so it should be set
+        # after validation
```
talledodiego, line 156: I totally agree on this and I like this very much!

structuralcodes/materials/concrete/_concreteEC2_2023.py
```diff
@@ -38,7 +36,29 @@ def __init__(
...
+                t.Literal[
+                    'elastic',
+                    'parabolarectangle',
+                    'bilinearcompression',
+                    'sargin',
+                    'popovics',
+                ],
+                ConstitutiveLaw,
+            ]
+        ] = 'parabolarectangle',
```
talledodiego, line 50: I like this in principle. And you ae right it gives flexibility. Maybe too much?
That was the idea in having stored in the constitutive law an attribute telling what materials can use that constitutive law. That is to avoid for instance setting a ParabolaRectangle law for reinforcement ... Do you think we should worry about this?

structuralcodes/materials/constitutive_laws/__init__.py
```diff
@@ -61,24 +61,25 @@ def create_constitutive_law(
...
-                f'Constitutive law {constitutive_law_name} not available for'
-                f' material {material.__class__.__name__}'
-            )
-    else:
-        raise ValueError(f'Unknown constitutive law: {constitutive_law_name}')
-    return law
+            # get the kwargs from the special dunder method
+            kwargs = method()
+            # create and return the constitutive law
+            return const_law(**kwargs)
```
talledodiego, line 80: Better than my original code! Thanks!

structuralcodes/materials/reinforcement/_reinforcementEC2_2004.py
```diff
@@ -20,6 +21,16 @@ def __init__(
...
+        constitutive_law: t.Optional[
+            t.Union[
+                t.Literal[
+                    'elastic',
+                    'elasticperfectlyplastic',
+                    'elasticplastic',
+                ],
+                ConstitutiveLaw,
+            ]
+        ] = 'elasticplastic',
```
talledodiego, line 33: Same comment as before on passing ConstitutiveLaw.

What I would like for extra flexibility is add the possibility have a GenericMaterial that accepts any ConstitutiveLaw. We could also create more specialized materials like `ElasticPlasticMaterial` etc. that used the proper constitutive laws with the given parameters.

structuralcodes/materials/concrete/_concreteEC2_2023.py
```diff
@@ -55,123 +75,213 @@ def __init__(
...
-        Returns:
-            float: The mean compressive strength in MPa.
-        """
-        self._fcm = self._fcm or ec2_2023.fcm(self.fck)
-        return self._fcm
+        # The constitutive law requires valid attributes, so it should be set
+        # after validation
+        self._constitutive_law = (
+            constitutive_law
+            if isinstance(constitutive_law, ConstitutiveLaw)
```
talledodiego, line 140: I am missing here a piece I think. In this way we are not checking the ConstitutiveLaw is valid for Concrete material, right? (see previous comment)

tests/test_materials/test_concrete.py
```diff
@@ -39,48 +38,65 @@ def test_constitutive_law_setter_factory():
...
         )
 
-        # Act and assert
-        concrete.constitutive_law = 'sargin'
-        assert isinstance(concrete.constitutive_law, Sargin)
-        assert math.isclose(concrete.constitutive_law._fc, -concrete.fcd())
+        # Act and assert for Sargin law
+        new_concrete = type(concrete)(
+            fck=concrete.fck, constitutive_law='sargin'
+        )
```
talledodiego, line 72: omg! so cool using type(concrete) for creating a new object of the same type!

talledodiego (changes requested):
Excellent job!
Some very few comments, if you want to consider them.

structuralcodes/materials/concrete/_concreteEC2_2004.py
mortenengen: The most compact way is perhaps to accept a `**kwargs`, and look for keywords inside that dict. However, in my opinion, that approach makes the signature of the `__init__` less transparent. I would rather prefer having the long, but explicit, signature. We could see how this evolves over time and later make it more compact if needed.

structuralcodes/materials/concrete/_concreteEC2_2023.py
mortenengen: I do not think we should worry about this, We could however, be more specific in the docstrings of the constitutive laws stating what materials the different laws are intended for. What do you think?

structuralcodes/materials/reinforcement/_reinforcementEC2_2004.py
mortenengen: I like the idea of having specialized material classes, but let us save that for a future PR 😃

structuralcodes/materials/concrete/_concreteEC2_2023.py
mortenengen: I intentionally skipped that check. If we wish to re-implement it, we should here check the materials of the `ConstitutiveLaw` object passed to `__init__`. The `create_constitutive_law` factory already raises an exception if the relevant magic method is not found on the material object. What do you think?

structuralcodes/materials/concrete/_concreteEC2_2004.py
talledodiego: I agree that I like more the explicit way.. Even if a very long `__init__` is not practical.
For now let's keep it like that.
Maybe in the future if it grows a lot we could create small dataclasses dividing the inputs in groups (for instance those related to constitutive laws - eps_c1, es_c2 etc. - those related to tensile behavior - fctm, fctk etc. - )...

structuralcodes/materials/reinforcement/_reinforcementEC2_2004.py
talledodiego: Yes, let's keep that for a future PR. what is related to this PR though is the possibility of passing "any" `ConstitutiveLaw` object. Should we constraint more the choice?

structuralcodes/materials/concrete/_concreteEC2_2023.py
talledodiego: The idea of the `__materials__` attribute in the constitutive laws is to avoid the use of wrong constitutive laws for the materials. For instance it could be a real mess use `BilinearCompression` with a `Reinforcement` material.
For the purpose of "flexibility" we created the `UserDefined` constitutive laws that can be used on any material giving the user complete responsibility on this action.

So I think it is dangerous to pass here a `ConstitutiveLaw` without checking if that constituve law is usable by that material (as discussed in the next unresolved comment). 

Maybe it is better to pass a `UserDefined`  constitutive law if we want to pass something for user flexibility? What do you think?

structuralcodes/materials/concrete/_concreteEC2_2023.py
talledodiego: Also related to previous comment...

structuralcodes/materials/concrete/_concreteEC2_2023.py
mortenengen: Got it. I have added validation in b221672 😃

talledodiego: excellent. Thanks!

talledodiego (approved):
Perfect job!
We have immutable materials now! Thanks!

PR #227, opened 2025-03-18
feat: implement shell section

### LINKED ISSUE #225: Add functionality for analysing shell sections
# Description

This is an attempt to describe the necessary steps for adding functionality for analyzing shell sections. The objective is to implement a `ShellSection` class with functionality similar to the current `GenericSection`, but capable of calculating the response of a shell with an arbitrary thickness and material, and with arbitrary reinforcement.

A shell section is subjected to the stress resultants shown in the figure below. As a start we should implement functionality for calculating the response of the moments and membrane forces with respect to the respective strain components relative to the mid-plane, i.e. three in-plane strain components, and three curvature components.

Image

The relations between stresses and stress resultants are given in the equations below, positive stresses are tension, and a positive curvature gives tension on the face of the shell with negative z-coordinate.

```math
n_{\textrm{x}} = \int_{-t/2}^{t/2} \sigma_{\textrm{x}} dz
```

```math
n_{\textrm{y}} = \int_{-t/2}^{t/2} \sigma_{\textrm{y}} dz
```

```math
n_{\textrm{xy}} = \int_{-t/2}^{t/2} \tau_{\textrm{xy}} dz
```

```math
m_{\textrm{x}} = -\int_{-t/2}^{t/2} \sigma_{\textrm{x}} z dz
```

```math
m_{\textrm{y}} = -\int_{-t/2}^{t/2} \sigma_{\textrm{y}} z dz
```

```math
m_{\textrm{xy}} = -\int_{-t/2}^{t/2} \tau_{\textrm{xy}} z dz
```

The integration of stresses to stress resultants should be performed by dividing the section into _n_ layers, or fibers through the thickness, where each layers is assumed to be in a state of plane stress.

# Comparison between `GenericSection` and `ShellSection`

The overview below compares the current `GenericSection` and the proposed `ShellSection` and outlines tasks that should be completed.

### Material

Concrete:

- Uniaxial stress-strain diagram

Reinforcement:

- Uniaxial stress-strain diagram.

### Gross geometry

- Surface represented as an arbitrary polygon. Wrappers for creating rectangle and circle.
- [ ]  Points to a material. Today, we point to a constitutive law. This will be updated.
- Leverages `_repr_svg_` for visualizing the geometry.

### Reinforcement

- `Point`
- [ ]  Points to a material. Today, we point to a constitutive law. This will be updated.
- Reinforcement is added to a geometry to create a `CompoundGeometry` using the following functions:
`add_reinforcement`
`add_reinforcement_line`
`add_reinforcement_circle`

### Section

- [ ]  `GenericSection` should be considered renamed to `GenericBeamSection`.

### SectionCalculator

- [ ]  `GenericSectionCalculator` should be considered renamed to `GenericBeamSectionCalculator`.
- [ ]  With the new `ShellSectionCalculator`, the abstract methods `calculate_bending_strength` and `calculate_moment_curvature` on the base class `SectionCalculator` are no longer well defined. These should be removed.

### Integrator

- [ ]  We should consider renaming the IntegratorFactory and all related to e.g. `BeamIntegratorFactory`.

### Material

Concrete

- Uniaxial stress-strain diagram. As before. However, we should probably implement a `.get_stress_2d` method that accepts strains from a state of plane stress, and returns the corresponding stresses.
- [ ]  Reduction of compressive strength due to transverse tension.
- [ ]  Increase of compressive strength due to transverse compression.

Reinforcement

- Uniaxial stress-strain diagram. As before.

### Gross geometry

- [x]  A shell represented by it’s thickness, e.g.: `ShellGeometry(thickness=450, material=concrete)`
- [x]  It should inherit from `Geometry`.
- [x]  Points to a material.
- [x]  It should have an attribute, e.g. `.reinforcement`, which could be a list of reinforcement objects.
- [x]  Could implement a `_repr_svg_` which visualizes the reinforcement in the plane.

### Reinforcement

- [x]  A custom `ShellReinforcement` object with at least the following attributes:
    - The position over the thickness.
    - The number of bars in a bundle.
    - The center distance between bar bundles.
    - The bar diameter.
    - The material.
    - The orientation in the plane.
- [ ]  Could be added to a shell geometry with for example a function `add_shell_reinforcement`. This function could act as a factory function which takes the required arguments for creating a `ShellReinforcement` object, create it, and store it in the `.reinforcement` attribute of the `ShellGeometry` object.

### Section

- [ ]  `ShellSection` which stores a `ShellGeometry` and initializes and stores a `ShellSectionCalculator`. Inherits from `Section`.

### SectionCalculator

- [ ]  ShellSectionCalculator which sets attributes for an integrator and creates a `ShellIntegrator`. Inherits from `SectionCalculator`. Should have the following methods as a start:
    - `.integrate_strain_profile(strain: ArrayLike, integrate: t.Literal[’stress’, ‘modulus’] = ‘stress’) → t.Union[t.Tuple[float, float, float], NDArray]`
    - `.calculate_strain_profile(nx, ny, nxy, mx, my, mxy) → t.List[float]`

### Integrator

- [ ]  `ShellFiberIntegrator` with concrete implementations of the three abstract methods defined on `SectionIntegrator`.

mortenengen:
This PR tracks the development related to #225.


structuralcodes/materials/constitutive_laws/_elastic_2d.py
```diff
@@ -0,0 +1,86 @@
...
+        name: t.Optional[str] = None,
+    ) -> None:
+        """Initialize an Elastic2D Material.
+
+        Arguments:
+            E (float): The elastic modulus.
+            nu (float): Poisson's ratio.
+            name (str, optional): A descriptive name for the constitutive law.
+        """
+        super().__init__(E=E, name=name)
```
talledodiego: Just a small note for here and similar points: should we adapt these lines to the new default names convention?

structuralcodes/materials/constitutive_laws/_elastic_2d.py
```diff
@@ -0,0 +1,86 @@
...
+        """Return Poisson's ratio.
+
+        Note:
+        Including this property allows future checks or transformations
+        (e.g., ensuring 0 ≤ nu < 0.5).
+        """
+        return self._nu
+
+    @property
+    def C_matrix(self) -> np.ndarray:
```
talledodiego: Just a feeling: isn't this named D generally in literature?

structuralcodes/materials/constitutive_laws/_elastic_2d.py
```diff
@@ -0,0 +1,86 @@
...
+                    [
+                        [1.0, nu, 0.0],
+                        [nu, 1.0, 0.0],
+                        [0.0, 0.0, (1 - nu) / 2.0],
+                    ]
+                )
+            )
+        return self._C_matrix
+
+    def get_stress(self, eps: ArrayLike) -> np.ndarray:
```
talledodiego, line 67: this is different than our 1d version where eps can be an array for plotting for instance entire constitutive laws or passing a bunch of points. Should we aim at the same here? Maybe we could pass a n x 3 array? Or a 3 x n? I am fine with boths, we just must be clear about what we expect from the user.
But supposing we are giving a n x 3 the return should be simply `(self.C_matrix @ eps.T).T`

structuralcodes/materials/constitutive_laws/_concrete_smeared_cracking/_core.py
```diff
@@ -0,0 +1,185 @@
...
+from ._strength_reduction_lateral_cracking import (
+    StrengthReductionLateralCracking,
+)
+from ._utils import (
+    calculate_principal_strains,
+    establish_strain_transformation_matrix,
+)
+
+
+class ConcreteSmearedCracking:
```
talledodiego, line 23: This makes me wonder: should the base classes for constitutive laws 1D be also inside `constitutive_laws` and not `core` by analogy? A matter for future thinking I guess.

structuralcodes/materials/constitutive_laws/_elastic_2d.py
```diff
@@ -0,0 +1,86 @@
+import typing as t
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+from ._elastic import Elastic
+
+
+class Elastic2D(Elastic):
```
talledodiego: Should be this derived from Elastic? 
Elastic is a 1D law... I find it a bit confusing here...

structuralcodes/materials/constitutive_laws/_elastic_2d.py
talledodiego: After talk with ME: shall we call this stiffness_matrix and the internal _stiffness_matrix?

structuralcodes/materials/constitutive_laws/_elastic_2d.py
talledodiego: After talk with ME we decided to remove derivation from Elastic right now. In the future we can think about having ConstitutiveLaw (just very abstract with few methods), ConstitutiveLaw1D (with basically the methods we have right now in ConstituveLaw except the most generic interfaces, counters, etc.) and ConstitutiveLaw2D from which we can derive the concrete constitutive law classes. In this way we can have for instance free methods (that can always be overriden) like get_secant etc.

structuralcodes/materials/constitutive_laws/_elastic_2d.py
talledodiego: If we remove the derivation from Elastic we should change accordingly this line?

structuralcodes/materials/constitutive_laws/_elastic_2d.py
talledodiego: After our talk: we should consider passing this arrays in analogy with 1d const laws for user experience that have a similar approach. When it is not easy to vectorize them (like in Smeared) we do the explicit for loops in structural codes so that the user does not to write those explicit loops when using the const law for plotting for instance.

1D plot:
`law.get_stress(strains)`

2D current plot:
`[law.get_stress(strain for strain in strains)]`

structuralcodes/materials/constitutive_laws/_concrete_smeared_cracking/_core.py
talledodiego: After talk we agreed that this door is open for future PR.

structuralcodes/materials/constitutive_laws/_concrete_smeared_cracking/_core.py
```diff
@@ -0,0 +1,185 @@
...
+    _initial_modulus_compression: t.Optional[float] = None
+    _initial_modulus_tension: t.Optional[float] = None
+    _uniaxial_compression: ConstitutiveLaw
+    _strength_reduction_lateral_cracking: StrengthReductionLateralCracking
+    _poisson_reduction: PoissonReduction
+    _cracking_criterion: CrackingCriterion
+
+    def __init__(
+        self,
+        uniaxial_compression: ConstitutiveLaw,
```
talledodiego: should this be a uniaxial_compression or uniaxial_law? The pros and cons to think about are related also to the use of creacking_criterion vs uniaxial_tension... the constitutive laws in tension could have a __ method for saying when cracked. Something to think about. We can think about it also later even if it means a possible breaking change.

structuralcodes/materials/constitutive_laws/_concrete_smeared_cracking/_core.py
```diff
@@ -0,0 +1,185 @@
...
+        # If the assumption of uncracked holds after calculating effective
+        # strains, return the effective uncracked principal strains
+        if not self.cracking_criterion.cracked(eps_p=eps_pf_uncracked):
+            return eps_pf_uncracked
+
+        # Else, return the effective principal strains assuming cracked
+        return self.poisson_reduction.poisson_matrix(cracked=True) @ eps_p
+
+    @property
+    def uniaxial_compression(self) -> ConstitutiveLaw:
```
talledodiego: eventually change this in uniaxial_law (depending on what we decide above)

structuralcodes/sections/_shell_section.py
```diff
@@ -0,0 +1,249 @@
...
+        **kwargs,
+    ) -> None:
+        """Initialize the ShellSectionCalculator.
+
+        Arguments:
+            section (ShellSection): The section object.
+        """
+        super().__init__(section=section)
+        self.integrator = ShellFiberIntegrator()
+        self.n_layers = kwargs.get('n_layers', 100)
```
talledodiego: Are 100 layers too much as default case for the majority of uses?

structuralcodes/sections/section_integrators/__init__.py
```diff
@@ -6,11 +6,13 @@
 from ._fiber_integrator import FiberIntegrator
 from ._marin_integrator import MarinIntegrator
 from ._section_integrator import SectionIntegrator
+from ._shell_integrator import ShellFiberIntegrator
 
 __all__ = [
     'integrator_factory',
```
talledodiego, line 12: Now the integrator_factory is only for beams basically, right? Should we think about some renaming / refactory here?

talledodiego (commented):
Thanks so much for this outstanding contribution. Just some minor comments/questions. Eventually for future PR (shall we in case document some future work in proper issues?)

structuralcodes/materials/constitutive_laws/_elastic_2d.py
mortenengen: I agree that we should adapt to the default name convention, and this is easily carried out when we establish the class hierarchy as discussed. I have made a note of this in #374.

structuralcodes/materials/constitutive_laws/_elastic_2d.py
mortenengen: I agree, but I think both C and D can be found in the literature, better to rename to `stiffness_matrix` to avoid confusion 😄 solved in 3bbddee.

structuralcodes/materials/constitutive_laws/_elastic_2d.py
mortenengen: I agree. I have described this in #375 where we can track further development.

structuralcodes/materials/constitutive_laws/_concrete_smeared_cracking/_core.py
mortenengen: Good point. It is now mentioned in #374.

structuralcodes/materials/constitutive_laws/_elastic_2d.py
mortenengen: I totally agree. This is fixed in ecaf927 and mentioned in #374.

structuralcodes/materials/constitutive_laws/_concrete_smeared_cracking/_core.py
mortenengen: I agree, and it is renamed in b145cfb.

structuralcodes/sections/section_integrators/__init__.py
mortenengen: For now, I think it is sufficient to leave it like it is since there is only one integrator for the shell sections. We could consider renaming the factory to e.g. `beam_integrator_factory`, though?

structuralcodes/materials/constitutive_laws/_concrete_smeared_cracking/_core.py
mortenengen: Changed in b145cfb.

structuralcodes/sections/_shell_section.py
mortenengen: I totally agree. This is changed in 1530447.

PR #228, opened 2025-03-24
feat: implement shell geometry

krkris:


mortenengen (approved):
Thanks for the nice contribution 👍

PR #235, opened 2025-04-07
feat: add svg representation of shell reinforcement

saraschei:


structuralcodes/geometry/_shell_geometry.py
```diff
@@ -122,4 +124,110 @@ def add_reinforcement(
...
+                    px, py = -s, c
+                    # Color based on orientation
+                    col = (
+                        'red'
+                        if np.isclose(phi % np.pi, 0)
+                        else 'blue'
+                        if np.isclose(phi % np.pi, np.pi / 2)
+                        else 'green'
+                    )
+                    n = int(w / sp) + 3
```
mortenengen, line 154: Can you please explain why we need the `+ 3`?

mortenengen (changes requested):
Great job on this one. Please consider the comments while finalizing.

structuralcodes/geometry/_shell_geometry.py
saraschei: Adding +3 is a safety margin: +1 replaces the line lost to rounding down number of rebars (int), while the other +2 put one extra line beyond each edge. Those outer lines are clipped away, but they ensure no visible gaps when rotating the rebars.

mortenengen: Makes sense. Thanks for clarifying! While finalizing, you might consider adding a comment in the code which states this for the curious code reader 😃

structuralcodes/geometry/_shell_geometry.py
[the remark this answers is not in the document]
saraschei: The figure is now scaled by 0.6 to make it fit most screens. 

add_reinforcement() now throws an error if z lies outside the thickness, preventing bars outside the concrete.

structuralcodes/geometry/_shell_geometry.py
```diff
@@ -122,6 +122,21 @@ def add_reinforcement(
...
+                -half_thickness + r.diameter_bar / 2
+                <= r.z
+                <= half_thickness - r.diameter_bar / 2
+            ):
+                raise ValueError(
+                    f'Reinforcement at z = {r.z:.2f} mm is outside the'
+                    f'range [-{half_thickness:.2f}, {half_thickness:.2f}] mm.'
+                )
+            return self
+        return None
```
mortenengen, line 138: The two `return`s seem redundant. Please remove.

saraschei: Removed the lines in the latest commit

structuralcodes/geometry/_shell_geometry.py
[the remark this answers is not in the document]
mortenengen: I will put the side views on our to do list :smiley:

mortenengen (approved):
Good job. Thanks for the contribution!

PR #240, opened 2025-04-29
feat: add shell section calculator and tests

saraschei:
Adding ShellSection and ShellSectionCalculator with tests.

### DIFF
--- a/structuralcodes/sections/_shell_section.py
+++ b/structuralcodes/sections/_shell_section.py
@@ -192,11 +192,7 @@ class ShellSectionCalculator(SectionCalculator):
         while True:
             # Check if number of iterations exceeds the maximum
             if num_iter > max_iter:
-                break
-
-            # Calculate response and residuals
-            response = np.array(self.integrate_strain_profile(strain=strain))
-            residual = loads - response
+                raise StopIteration('Maximum number of iterations reached.')
 
             if initial:
                 # If the initial stiffness is used, we follow a regular
@@ -238,8 +234,5 @@ class ShellSectionCalculator(SectionCalculator):
             # Check for convergence:
             if np.linalg.norm(delta_strain) < tol:
                 break
-
-        if num_iter >= max_iter:
-            raise StopIteration('Maximum number of iterations reached.')
 
         return strain.tolist()
--- a/structuralcodes/sections/_shell_section.py
+++ b/structuralcodes/sections/_shell_section.py
@@ -125,6 +125,11 @@ class ShellSectionCalculator(SectionCalculator):
             self.layers = result[-1]
 
         # Return the results without layers
+        if len(result) == 2:
+            # Return only the stiffness matrix
+            return result[0]
+
+        # Return only the stress resultants
         return result[:-1]
 
     def calculate_strain_profile(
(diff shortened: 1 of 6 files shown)


structuralcodes/sections/_shell_section.py
```diff
@@ -52,6 +126,35 @@ def calculate_strain_profile(
...
+        integrator = self.integrator
+
+        strain = np.zeros(6)
+        prepared_input, _ = integrator.prepare_input(
+            geo=geo, strain=strain, integrate='modulus'
+        )
+        K = integrator.integrate_modulus(prepared_input)
+
+        # Solve for the strain profile using the pseudoinverse
+        return np.linalg.pinv(K) @ stress
```
mortenengen, line 160: Note that the direct solution of the equation will only work if the response is linear. As soon as we add materials with non-linear constitutive laws, we need to solve for the strains iteratively. Let's get back to this when dealing with the non-linear materials.

tests/test_sections/test_shell_section.py
```diff
@@ -0,0 +1,150 @@
...
+)
+def test_integrate_strain_profile(
+    Ec, nu, thickness, eps_x, eps_y, eps_xy, chi_x, chi_y, chi_xy
+):
+    """Elastic plate: strains → stress-resultants."""
+    material = Elastic2D(Ec, nu)
+    shell = ShellSection(ShellGeometry(thickness, material))
+
+    # Numerically integrated forces/moments
+    R = shell._section_calculator.integrate_strain_profile(
```
mortenengen, line 37: After adding the property `section_calculator`, see the comment above, this line should be updated to avoid accessing private attributes.

mortenengen (changes requested):
Thanks for this good contribution! I have left a couple of comments for you to consider while finalizing.

structuralcodes/sections/_shell_section.py
mortenengen: I think we should consider an implementation similar to the one from the [`GenericSectionCalculator`](https://github.com/fib-international/structuralcodes/blob/ba13326583edf8fbf96c21f097c55d889ea425b4/structuralcodes/sections/_generic.py#L1346).

structuralcodes/sections/_shell_section.py
```diff
@@ -1,25 +1,46 @@
...
+            The ShellSection uses a ShellSectionCalculator for all
+            calculations. The ShellSectionCalculator uses a
+            ShellFiberIntegrator for integrating over the thickness.
+        """
+        if name is None:
+            name = 'ShellSection'
+        super().__init__(name)
         self._geometry = geometry
-        self.section_calculator = ShellSectionCalculator(section=self)
+        self._section_calculator = ShellSectionCalculator(
```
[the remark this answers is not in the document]
mortenengen: `section_calculator` was made public in b3a8d1a.

tests/test_sections/test_shell_section.py
mortenengen: The tests are updated in b3a8d1a and 7a71ece.

mortenengen (approved):
Thanks for the good contribution and nice collaboration on this one 😃

PR #242, opened 2025-05-06
feat: add ParabolaRectangle2D material class with tests

krkris:


structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -0,0 +1,159 @@
...
+        """Return the second coefficient for the compressive-strength reduction
+        factor due to lateral tension. Default value = 100.
+        """
+        return self._c_2
+
+    def transform(self, strain: ArrayLike) -> np.ndarray:
+        """Transform the strain vector to principal directions."""
+        eps = np.atleast_1d(strain)
+        # Use arctan2 to obtain the angle in the correct quadrant
+        theta = 0.5 * np.arctan2(2 * eps[2], eps[0] - eps[1])
```
mortenengen, line 72: Since `eps[2]` is gamma_xy, and not epsilon_xy, the expression on line 72 should be, i.e. skipping the factor `2`:

```python
theta = 0.5 * np.arctan2(eps[2], eps[0] - eps[1])
```

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -0,0 +1,159 @@
...
+        c = np.cos(theta)
+        s = np.sin(theta)
+        T = np.array(
+            [
+                [c * c, s * s, s * c],
+                [s * s, c * c, -s * c],
+                [-2 * s * c, 2 * s * c, c * c - s * s],
```
mortenengen, line 79: To yield consistent transformations, the final row of the transformation matrix should be:

```python
[2 * s * c, -2 * s * c, s * s - c * c],
```

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -0,0 +1,159 @@
...
+    def get_effective_principal_strains(
+        self, eps_p: ArrayLike, nu: float
+    ) -> np.ndarray:
+        """Compute the effective principal strains to include the influence of
+        Poisson's ratio. Taken from 'Nonlinear Analysis of Reinforced-Concrete
+        Shells' by M. A. Polak and F. J. Vecchio (1993).
+        """
+        denom = (1 + nu) * (1 - 2 * nu)
+
+        P = np.array(
```
mortenengen, line 93: Please note that `eps_p` should be an array with two positions holding the two principal strains. Therefore we only need to consider `P` as a 2x2 matrix.

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -0,0 +1,159 @@
...
+
+        sig_p = super().get_stress(eps_pf)
+        sig_p[2] = 0
+
+        for i in (0, 1):
+            if sig_p[i] < 0:
+                sig_p[i] *= beta
+
+        # Transform back to global coords
+        return T.T @ sig_p
```
mortenengen, line 140: Ref. the comment above. `eps_p` and `eps_pf` should be arrays with two positions, and this would result in `sig_p` being an array with two positions. If this change is implemented, line 133 should be skipped, and line 140 should be slightly modified to:

```python
return T.T @ np.array([*sig_p, 0])
```

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -0,0 +1,159 @@
...
+
+        return min(beta, 1)
+
+    def get_stress(self, strain: ArrayLike) -> np.ndarray:
+        """Return a 2D stress vector [sigma_x, sigma_y, tau_xy]
+        given a 2D strain vector [eps_x, epx_y, gamma_xy].
+        """
+        T, eps_p = self.transform(strain)
+
+        cracked = np.any(eps_p > 0)
```
mortenengen, line 124: Ref. the comment above. Since `eps_p[2]` for the moment holds the shear strain in the principal direction, which might be different from zero due to round off, and possibly slightly larger than zero, this check might not be correct in all cases. Please consider modifying it to e.g.:

```python
cracked = np.any(eps_p[:2] > 0)
```

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -0,0 +1,159 @@
...
+        return T.T @ sig_p
+
+    def get_secant(self, strain: ArrayLike) -> np.ndarray:
+        """Compute the 3x3 secant stiffness matrix C."""
+        T, eps_p = self.transform(strain)
+        cracked = np.any(eps_p > 0)
+        nu = 0.0 if cracked else self._nu
+        eps_pf = self.get_effective_principal_strains(eps_p, nu)
+
+        sig_p = super().get_stress(eps_pf)
```
mortenengen, line 149: Ref. comments above about the length of `eps_p`.

mortenengen (changes requested):
Thanks for the good contribution! I have left a couple of comments for you to consider while finalizing the code 😃

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
krkris: The factor `2` is removed in 35898f8

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
krkris: I have added the `check_cracked` method as you suggested. Ref. the comments above, `eps_p` is now an array with two positions, so it is not necessary to slice the list in the check. This is fixed in I 35898f8.

krkris:
The computation of the elastic moduli is now refined to avoid division by zero. The tolerance is set to 1e-12 for now. This is updated in 404ce4c.

mortenengen (approved):
Thanks for the good contribution and for implementing the fixes!

Please note that in the current implementation, the secant stiffness matrix is not fully consistent since we are not taking into account the coupling terms due to poisson effects. Let us leave it like this for now, and revisit it later to improve the convergence rate 👍

PR #243, opened 2025-05-07
feat: add ElasticPlastic2D material class with tests

saraschei:


tests/test_materials/test_elasticplastid_2d.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
```
mortenengen, line 1: Ruff is complaining because there is no docstring in this file. Please add one 😃

saraschei: Fixed in commit 5322332.

structuralcodes/materials/constitutive_laws/__init__.py
```diff
@@ -7,7 +7,9 @@
 from ._elastic import Elastic
 from ._elastic_2d import Elastic2D
 from ._elasticplastic import ElasticPlastic
+from ._elasticplastic_2d import ElasticPlastic2D
 from ._parabolarectangle import ParabolaRectangle
+from ._parabolarectangle_2d import ParabolaRectangle2D
```
mortenengen, line 12: Note that the tests are failing because we are trying to import `ParabolaRectangle2D`, but this is not available, since it is being implemented in #242. Please remove this for now, and we will make sure it exists after merging #242.

saraschei: Removed in commit 3362924

PR #244, opened 2025-05-08
feat: add side view to _repr_svg in ShellGeometry

saraschei:


mortenengen (approved):
Nice! Thanks for adding this feature 😃

PR #245, opened 2025-05-08
feat: add tests for ShellGeometry and ShellReinforcement

krkris:


structuralcodes/geometry/__init__.py
```diff
@@ -30,8 +31,11 @@
     'IPN',
     'UPN',
     'add_reinforcement',
+    'add_reinforcement',
```
mortenengen, line 34: Please remove the duplicate line.

mortenengen (changes requested):
Thanks for the contribution! I have added one minor comment for you to consider while finalizing.

structuralcodes/geometry/__init__.py
krkris: Fixed in 2c38859

mortenengen (approved):
Good job! Thanks for the good contribution 😃

PR #246, opened 2025-05-12
feat: add stiffness reinforcement

saraschei:


structuralcodes/sections/section_integrators/_shell_integrator.py
```diff
@@ -124,17 +111,42 @@ def integrate_modulus(
...
+                    [c * c, s * s, c * s],
+                    [s * s, c * c, -c * s],
+                    [-2 * c * s, 2 * c * s, c * c - s * s],
+                ]
+            )
+
+            Cs_global = T.T @ Cs_local @ T
+
+            As = r.n_bars * np.pi * (r.diameter_bar / 2) ** 2
+            z = r.z
```
mortenengen, line 140: Please note that these lines should be performed in the `prepare_input` method, in such a way that the for-loop on lines 117-120 also cover the reinforcement.

structuralcodes/sections/section_integrators/_shell_integrator.py
```diff
@@ -124,17 +111,42 @@ def integrate_modulus(
...
+            Ksr = np.zeros((6, 6))
+            Ksr[:3, :3] = As * Cs_global
+            Ksr[:3, 3:] = -z * As * Cs_global
+            Ksr[3:, :3] = -z * As * Cs_global
+            Ksr[3:, 3:] = (z**2) * As * Cs_global
```
mortenengen, line 145: Please note how `As * Cs_global` er comparable to `C_dz` in lines 117-120, such that after the refactoring described above, line 142 will be covered by line 118, line 143 and 144 will be covered by line 119, and line 145 will be covered by line 120.

structuralcodes/sections/section_integrators/_shell_integrator.py
```diff
@@ -80,9 +71,7 @@ def prepare_input(
                 raise ValueError(f'Unknown integrate type: {integrate}')
 
             IA.append(integrand * dz)
-
         prepared_input = [(np.array(z_coords), np.array(IA))]
-
         return prepared_input, z_coords
 
     def integrate_stress(
```
mortenengen, line 77: For the moment, the `integrate_stress` method does not include the response in the reinforcement. When refactoring as suggested below, the `prepare_input` method should include the necessary data in `z` and `stress_resultants` to also include the reinforcement.

mortenengen (changes requested):
Thanks for the good first implementation. Please consider the comments and update 👍

structuralcodes/sections/section_integrators/_shell_integrator.py
saraschei: Taken care of in commit f23bfd5. Removed geometry argument and moved the loop into prepare_input.

structuralcodes/sections/section_integrators/_shell_integrator.py
saraschei: Taken care of in commit f23bfd5. Deleted separate stiffness matrix for reinforcement; the one from before takes care of both concrete and reinforcement.

structuralcodes/sections/section_integrators/_shell_integrator.py
saraschei: Taken care of in commit f23bfd5

structuralcodes/sections/section_integrators/_shell_integrator.py
```diff
@@ -80,8 +72,36 @@ def prepare_input(
...
+            material = r.material
+            z_r = r.z
+            c, s = np.cos(r.phi), np.sin(r.phi)
+            T = np.array(
+                [
+                    [c * c, s * s, c * s],
+                    [s * s, c * c, -c * s],
+                    [-2 * c * s, 2 * c * s, c * c - s * s],
+                ]
+            )
```
mortenengen, line 87: I am sorry, I should have spotted this in the first round. Since the output of the code in lines 80-87 does not change between iterations, it could be put in a property on the `ShellReinforcement` to avoid it being recalculated in each iteration. Please consider this while finalizing.

structuralcodes/sections/section_integrators/_shell_integrator.py
```diff
@@ -80,8 +72,36 @@ def prepare_input(
...
+                integrand = material.get_stress(fiber_strain)
+            elif integrate == 'modulus':
+                integrand = material.get_tangent(fiber_strain)
+            else:
+                raise ValueError(f'Unknown integrate type: {integrate}')
+
+            As = r.n_bars * np.pi * (r.diameter_bar / 2) ** 2
+
+            IA.append(T.T @ integrand @ T * As)
+            z_list.append(z_r)
```
mortenengen, line 100: Please note that the fiber strain should be transformed to the reinforcement direction, and that only the strain component in the reinforcement direction should be passed to the `get_stress` and the `get_secant` method.

The return value from both of these methods is a scalar, i.e. a `float`, however, the integrand to store in `MA` should be either an array of size 3 for `stress` or 3x3 for `modulus`.

For `stress`, you create an array with the longitudinal stress in position 0, and zeros elsewhere, multiply it with `As`, and transform it by pre-multiplying the the transformation matrix like [here](https://github.com/krkris/structuralcodes/blob/add-parabola-rectangle-2d/structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py#L132-L133).

For `modulus` you create an array with the stiffness in position (0,0), and zeros elsewhere, multiply it with `As` and transform it the way you already did in line 99.

You might consider doing the array formation and transformation inside the if block starting on line 90 since it differs slightly between the two options.

structuralcodes/sections/section_integrators/_shell_integrator.py
```diff
@@ -80,8 +72,36 @@ def prepare_input(
...
+                    [-2 * c * s, 2 * c * s, c * c - s * s],
+                ]
+            )
 
-        prepared_input = [(np.array(z_coords), np.array(IA))]
+            fiber_strain = strain[:3] + z_r * strain[3:]
+            if integrate == 'stress':
+                integrand = material.get_stress(fiber_strain)
+            elif integrate == 'modulus':
+                integrand = material.get_tangent(fiber_strain)
```
mortenengen, line 93: Please use the `get_secant` method to be consistent with the treatment of the volume material.

mortenengen (changes requested):
Thanks for the improvement. I have a couple of comments that you should consider while finalizing 😃

structuralcodes/sections/section_integrators/_shell_integrator.py
saraschei: Moved line 80-87 into ShellReinforcement as a property in commit 37364b4b704.

structuralcodes/sections/section_integrators/_shell_integrator.py
saraschei: The strain is now transformed to the reinforcement direction and the values are handled as arrays. Taken care of in commit 3c933b3.

structuralcodes/sections/section_integrators/_shell_integrator.py
saraschei: Taken care of in commit 3c933b3

PR #248, opened 2025-05-14
fix: bugfix rotation triangulation data

talledodiego:
This commit fixes the bug for which results where inconsistent if evaluating the bending strength for an angle different than zero as the first computation on the section. Indeed the triangulation is created on the section without account the rotation of the section during computation of the strength.


mortenengen (approved):
Thanks for fixing this!

PR #249, opened 2025-05-15
fix: As calculation and updated notation for cc_bars

saraschei:


mortenengen (approved):
Thanks for fixing this 😃

PR #250, opened 2025-05-15
feat: use only material objects in geometry objects

mortenengen:
Removes the option to pass a constitutive law when creating a geometry. Instead, only material objects are allowed.

The following tasks were done:

- Allow only material objects when creating geometry objects.
- Add basic material objects with basic constitutive laws.
- Update tests.

### DIFF
--- a/structuralcodes/geometry/_geometry.py
+++ b/structuralcodes/geometry/_geometry.py
@@ -18,9 +18,9 @@ from shapely.geometry import (
 )
 from shapely.ops import split
 
-from structuralcodes.core.base import ConstitutiveLaw, Material
+from structuralcodes.core.base import Material
+from structuralcodes.materials.basic import ElasticMaterial
 from structuralcodes.materials.concrete import Concrete
-from structuralcodes.materials.constitutive_laws import Elastic
 
 
 class Geometry:
@@ -72,7 +72,7 @@ class Geometry:
     @staticmethod
     def from_geometry(
         geo: Geometry,
-        new_material: t.Optional[t.Union[Material, ConstitutiveLaw]] = None,
+        new_material: t.Optional[Material] = None,
     ) -> Geometry:
         """Create a new geometry with a different material."""
         raise NotImplementedError(
@@ -91,8 +91,7 @@ class PointGeometry(Geometry):
         self,
         point: t.Union[Point, ArrayLike],
         diameter: float,
-        material: t.Union[Material, ConstitutiveLaw],
-        density: t.Optional[float] = None,
+        material: Material,
         name: t.Optional[str] = None,
         group_label: t.Optional[str] = None,
     ):
@@ -105,12 +104,7 @@ class PointGeometry(Geometry):
             point (Union(Point, ArrayLike)): A couple of coordinates or a
                 shapely Point object.
             diameter (float): The diameter of the point.
-            material (Union(Material, ConstitutiveLaw)): The material for the
-                point (this can be a Material or a ConstitutiveLaw).
-            density (Optional(float)): When a ConstitutiveLaw is passed as
-                material, the density can be providen by this argument. When
-                the material is a Material object the density is taken from the
-                material.
+            material (Material): The material for the point.
             name (Optional(str)): The name to be given to the object.
             group_label (Optional(str)): A label for grouping several objects
                 (default is None).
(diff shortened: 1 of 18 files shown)


structuralcodes/geometry/_geometry.py
```diff
@@ -698,20 +656,23 @@ class CompoundGeometry(Geometry):
...
-    geometries: t.List[Geometry]
+    geometries: t.List[t.Union[SurfaceGeometry, PointGeometry]]
 
     def __init__(
         self,
-        geometries: t.Union[t.List[Geometry], MultiPolygon],
+        geometries: t.Union[
+            t.List[t.Union[SurfaceGeometry, PointGeometry, CompoundGeometry]],
+            MultiPolygon,
+        ],
```
talledodiego, line 666: Correct, I was wrong in my original implementation! Thanks for having fixed this!

structuralcodes/materials/__init__.py
```diff
@@ -1,9 +1,10 @@
 """Main entry point for materials."""
 
-from . import concrete, constitutive_laws, reinforcement
+from . import basic, concrete, constitutive_laws, reinforcement
 
 __all__ = [
     'concrete',
     'constitutive_laws',
     'reinforcement',
+    'basic',
```
talledodiego, line 9: I am not sure on the name `basic`, perhaps `standard`?

structuralcodes/materials/basic/_elastic.py
```diff
@@ -0,0 +1,38 @@
+"""A material class with elastic properties."""
```
talledodiego, line 1: In this file a would create a static class method that creates an elasticmaterial from a given material as argument, makes sense?

structuralcodes/materials/basic/_generic.py
```diff
@@ -0,0 +1,26 @@
+"""A generic material that could hold any type of constitutive law."""
```
talledodiego, line 1: mmm, now I realize that I like more the `standard` name for elastic and elastic plastic materials. For generic I don't like it so much...

structuralcodes/sections/_rc_utils.py
```diff
@@ -57,13 +58,21 @@ def create_surface_geometries(polygons_list, material):
...
     for geo in rotated_geometry.geometries:
-        Ec = geo.material.get_tangent(eps=0)
-        elastic_concrete = UserDefined([-100, 0], [-100 * Ec, 0])
+        Ec = geo.material.constitutive_law.get_tangent(eps=0)
+        density = geo.material.density
+        elastic_concrete_law = UserDefined([-100, 0], [-100 * Ec, 0])
+        elastic_concrete = GenericMaterial(
+            density=density,
+            constitutive_law=elastic_concrete_law,
+        )
```
talledodiego, line 67:
```suggestion
        elastic_concrete = GenericMaterial(
            density=density,
            constitutive_law=elastic_concrete_law,
            name='elastic concrete',
        )
```

talledodiego (approved):
Outstanding job!
There are just a couple of minor comments for your consideration.

structuralcodes/materials/__init__.py
mortenengen: Ref. our discussion on another platform. `basic` often refers to "the simplest parts", however `standard` often refers to "things which are usual and normal" or "a level that is thought to be acceptable". With this in mind, we agreed to go for `basic`.

structuralcodes/materials/basic/_generic.py
mortenengen: See the response to the comment above.

structuralcodes/geometry/_geometry.py
```diff
@@ -261,24 +242,24 @@ def from_geometry(
...
                     {repr(new_material)}'
                 )
         else:
             # new_material not provided, assume elastic material with same
             # elastic modulus
-            new_material = Elastic(E=geo.material.get_tangent(eps=0))
+            new_material = ElasticMaterial(
+                E=geo.material.constitutive_law.get_tangent(eps=0),
+                density=geo.material.density,
+            )
```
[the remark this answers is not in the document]
mortenengen: Thanks for the suggestion! I have added this in 37594a0.

structuralcodes/geometry/_geometry.py
```diff
@@ -618,22 +579,21 @@ def from_geometry(
...
```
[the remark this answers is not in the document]
mortenengen: Fixed in 37594a0.

structuralcodes/materials/basic/_elastic.py
mortenengen: Fixed in 37594a0.

structuralcodes/sections/_rc_utils.py
mortenengen: Thanks, fixed in 82c9b53

structuralcodes/materials/__init__.py
talledodiego: I agree. In my mind "things which are usual and normal" is the exact reason I would have preferred standard, but since there is also the other interpretation, we can leave as basic.

talledodiego (approved):
Excellent! thanks for this modification!

PR #252, opened 2025-05-19
fix: assign reinforcement as material in shell integrator

saraschei:


mortenengen (approved):
Thanks for providing this fix 😃

PR #253, opened 2025-05-19
feat: make consistent stiffness matrix for ParabolaRectangle2D

krkris:
This is an attempt of making a consistent secant stiffness matrix for the ParabolaRectangle2D material class.


structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -81,23 +81,24 @@ def transform(self, strain: ArrayLike) -> np.ndarray:
...
+        """Return the Poisson's ratio matrix."""
         denom = (1 + nu) * (1 - 2 * nu)
-
-        P = np.array(
+        return np.array(
             [
                 [(1 - nu) / denom, nu / denom],
                 [nu / denom, (1 - nu) / denom],
             ]
         )
```
mortenengen, line 92: Please note that these expressions are valid for the case of a general 3D state of stress. For plane stress, which is our case, the lines simplify to:

```python
...
denom = 1 - nu**2
return np.array(
    [1, nu],
    [nu, 1].
)
```

You can derive these expressions from Hooke's law for elastic materials for plane stress.

Please update the code accordingly.

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
```diff
@@ -149,25 +150,44 @@ def get_secant(self, strain: ArrayLike) -> np.ndarray:
...
-            E22 = (sig_p[1] * (beta if sig_p[1] < 0 else 1.0)) / eps_pf[1]
+            E_22 = (sig_p[1] * (beta if sig_p[1] < 0 else 1.0)) / eps_pf[1]
         else:
-            E22 = self._fc * 2.0 / self._eps_0
+            E_22 = self._fc * 2.0 / self._eps_0
 
-        E12 = (E11 + E22) / 2
+        # Initial 2x2 secant
+        D = np.diag([E_11, E_22])
+        C = P.T @ D @ P
```
mortenengen, line 177: Following the derivation of the poisson matrix from Hooke's law, the pre- AND post-multiplication is not necessary. The line should be simplified to:

```python
C = P @ D
```

Please update.

mortenengen (changes requested):
Thanks for the contribution. I have left a couple of comments that should be addressed while finalizing the code.

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
krkris: Thanks for pointing that out! The update is implemented in 2d0d87f.

PS: I assume you just forgot to multiply the matrix by `1 / denom` in the return expression.

structuralcodes/materials/constitutive_laws/_parabolarectangle_2d.py
krkris: This is updated in 2d0d87f.

mortenengen (approved):
Thanks for the contribution 😃

PR #255, opened 2025-05-22
EN-1992-1-1 2004 article6 improvement

MestreCarlos:
Thi add Equation 6.13, but isolating the reinforcement term, since it’s usually more common to want to determine the required reinforcement ratio


structuralcodes/codes/ec2_2004/shear.py
```diff
@@ -400,6 +400,51 @@ def VRds(
...
 
 
+# Equation (6.8 & 6.13) -> Reinforcement ratio isolated from the equation.
+def Asw_s_required(
+    Ved: float,
+    z: float,
+    theta: float,
+    fyk: float,
+    alpha: float = 90.0,
+    gamma_s: float = 1.15,
```
mortenengen: Please note that in Eq. (6.13), neither fyk nor gamma_s are input, but the derived valued fywd. Please update the implementation so that the function is one-to-one with the code equation.

MestreCarlos:
Done. Thanks for the review.

mortenengen (approved):
Thanks for the contribution 😃

PR #258, opened 2025-05-29
fix: replace misleading variable name in integrate_stress method

krkris:
In `integrate_stress`, the unpacking `z, stress_resultants = prepared_input[0]` might be misleading, since `stress_resultants` here is the array of per-fiber stress contributions from `prepare_input`.


mortenengen (approved):
Thanks for the improvement!

PR #259, opened 2025-06-02
fix: replace exponent factor in ParabolaRectangle2D

saraschei:


mortenengen (approved):
Thanks for fixing this 👍

PR #262, opened 2025-08-05
feat: Implement InitStrain

talledodiego:
This PR is an attempt for incorporating initial Strains and Stresses into constitutive laws.

The approach followed consisted in creating a wrapping constitutive law that stores the initial strain. The relevant methods used for getting material response (e.g. `get_stress`) are then passed to the wrapped material adding the initial strain.

For instance this is the stress-strain plot of a ElasticPlastic constitutive law wrapped with a InitStrain constitutive law:

```
prestress_reinf = create_reinforcement(fyk=1620, Es=200000, ftk=1620, epsuk=0.02)
init_strain_law = InitStrain(constitutive_law=prestress_reinf.constitutive_law, initial_strain=0.0054)

plt.plot(strain, init_strain_law.get_stress(strain), label='Prestressing Steel with Initial Strain')
strain = np.linspace(0, -0.025, 200)
plt.plot(strain, init_strain_law.get_stress(strain), label='Prestressing Steel with Initial Strain')
plt.axvline(0, color='black', linestyle='--', label='Zero Strain')
plt.xlabel('Strain')
plt.ylabel('Stress (MPa)')
```

This gives the following result:
image

### Section analysis with prestressed reinforcement
With this feature, it is possible to model prestressed section as the following example shows.

The example is taken from the document (in Italian): https://gelfi.unibs.it/software/Vcaslu-help.pdf

image

A initial stress of 1080 MPa is applied, corresponding to a initial strain of 5.4‰.

The moment capacity is 2273 kNm. The strain shown on prestressed reinforcement is 14.82‰ which includes the pre-strain; therefore the strain on the prestressing reinforcement is 14.8 - 5.4 = 9.4‰.

image

```
poly = Polygon([
    (250, 0), 
    (250, 140), 
    (70, 200), 
    (70, 920), 
    (250, 980), 
    (250, 1120), 
    (-250, 1120), 
    (-250, 980), 
    (-70, 920), 
    (-70, 200), 
    (-250, 140), 
    (-250, 0),
])

geo = SurfaceGeometry(poly=poly, material=concrete)
for y, n in zip([40, 110, 1010, 1080], [4, 2, 2, 4]):
    geo = add_reinforcement_line(
        geo=geo,
        coords_i=(-210, y), 
        coords_j=(210, y),
        diameter=10,
        n=n,
        material=reinforcement,
    )
area = 840
d = (area * 4 / np.pi)**0.5
geo = add_reinforcement_line(
    geo=geo,
    coords_i=(0, 70),
    coords_j=(0, 210),
    diameter=d,
    n=2,
    material=initStrain_mat
)
geo
```
image

```
section = GenericSection(geometry=geo, integrator='fiber', mesh_size=0.001)
res = section.section_calculator.calculate_bending_strength()
print(f'Bending strength: {res.m_y*1e-6:.2f} kNm')
```

The results obtained with structural codes are consistent with the example, with a bending strength of 2273.7 kNm and a strain in reinforcement of 9.4‰.

### TO BE DONE
The following modifications can be included:

- [x] Creation of a material (for now, as seen in the example, a `GenericMaterial` is used); this could be approached in a couple of possible ways: (1) creating a InitStrain material that wraps another material. This has some cons like for instance the access to the wrapped material properties with something like `material.base_material.fyd()` which is not optimal. (2) creating a `apply_initial_strain` to the base `Material` class: this function could always define a new constitutive law of type `InitStrain` passing its base constitutive law with the given pre-strain. In this way each material could have an initial strain to its constitutive law. Just pay attention to avoid cascade effects, like applying an initialStrain multiple times.
- [x] Creation of InitStress version since in some cases it may be preferred by the user.
- [x] Check that prestressed sections work in different conditions

### Final notes
This PR is an attempt for open discussion.
Another possible approach would be working at the level of the geometries, assigning a initial strain or stress to the geometry instead of the material. In this scenario the materials remain as they are, but the geometry adds the initial strain before passing it to the material. A complication of this approach could be retrieval of ultimate stresses on materials that do not account for initial strain.


structuralcodes/materials/constitutive_laws/_init_strain.py
```diff
@@ -0,0 +1,109 @@
...
+from __future__ import annotations  # To have clean hints of ArrayLike in docs
+
+import typing as t
+
+from numpy.typing import ArrayLike
+
+from ...core.base import ConstitutiveLaw
+
+
+class InitStrain(ConstitutiveLaw):
```
mortenengen: Really minor comment: could we call it `InitialStrain` to have the name one-to-one with the scope?

structuralcodes/materials/constitutive_laws/_init_strain.py
```diff
@@ -0,0 +1,109 @@
...
+
+    _wrapped_law: ConstitutiveLaw = None
+
+    def __init__(
+        self,
+        constitutive_law: ConstitutiveLaw,
+        initial_strain: float,
+        name: t.Optional[str] = None,
+    ) -> None:
+        """Initialize an Initial Strain Material.
```
mortenengen: The first line in the docstring should say `... Initial Strain Constitutive Law.`

structuralcodes/materials/constitutive_laws/_init_strain.py
```diff
@@ -0,0 +1,109 @@
...
+
+        This constitutive law is a wrapper for another constitutive law
+        that assigns an initial strain.
+
+        Arguments:
+            constitutive_law (ConstitutiveLaw): Wrapped constitutive law.
+            initial_strain (float): The initial strain to be applied.
+        """
+        name = name if name is not None else 'InitStrainLaw'
```
mortenengen: In line with the first minor comment, the default name could be `InitialStrainLaw`.

mortenengen (changes requested):
Thanks for the good draft contribution! Have only added a couple of really minor comments. Before merging, we should go ahead and stress test this feature as indicated in the todo of the PR.

structuralcodes/materials/constitutive_laws/_init_strain.py
talledodiego: Sure!

structuralcodes/materials/constitutive_laws/_init_strain.py
talledodiego: Done!

structuralcodes/materials/constitutive_laws/_init_strain.py
talledodiego: Done

PR #263, opened 2025-08-07
fix: accept kwargs in the reinforcement factory

mortenengen:
This fixes an inconsistency between the concrete factory and the reinforcement factory. `kwargs` like for example `constitutive_law` are now passed further to the specific reinforcement material.


mortenengen:
> Perfect. In the docstrings, it is customary to write something about kwargs?

Good question. In the factory for concretes we did not include anything, but for sure it would improve transparency to write a small note. I will do so in both factories before merging.

Thanks for noting this 👍

PR #265, opened 2025-08-08
fix: Update __sargin__  with k_sargin computed using fcd and not fcm

talledodiego:
This PR is a tentative solution for fixing the wrong initial stiffness read by `get_tangent(0.0)` method in `concrete`. 
The relation k_sargin (taken from (3.14) in EC2 was computed with reference to `fcm`. This PR makes this calculation with reference to fcd. The tests are broken now of course, but I will update them if we decide this could be the correct solution.

See the difference between the old (dashed orange line) and the current implementation (blue solid line).

image

This has ben corrected for EC2 2004 and EC2 2023. Not for MC2010 that is more tricky since gives values for given fck in a tabular format so making it harder deciding how to correct it when the law represents `fcd` instead of `fcm`.

image

### DIFF
--- a/structuralcodes/materials/concrete/_concreteEC2_2004.py
+++ b/structuralcodes/materials/concrete/_concreteEC2_2004.py
@@ -506,11 +506,19 @@ class ConcreteEC2_2004(Concrete):  # noqa: N801
 
     def __sargin__(self) -> dict:
         """Returns kwargs for creating a Sargin const law."""
+        if self._k_sargin is None:
+            k_sargin_design = ec2_2004.k_sargin(
+                Ecm=self.Ecm,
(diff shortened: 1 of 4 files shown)


talledodiego:
> Thanks for providing this tentative fix, @talledodiego, and for highlighting the effects.
> 
> A similar fix, i.e. producing a reasonable initial stiffness for a given strength class, could have been obtained if we passed `fcm` in `__sargin__` instead of `fcd()`. I am more in favor of that alternative instead of computing `k_sargin` with `fcd`, since passing `fcm` would be one-to-one with the equation in the design code. What do you think?

I agree with your idea. I will modify the code accordingly and submit for your review.

PR #266, opened 2025-08-20
docs: add usage guide

mortenengen:
This is a draft usage guide for the docs.


docs/usage/materials/index.md
```diff
@@ -0,0 +1,173 @@
...
+::::{dropdown-syntax}
+:::{literalinclude} ../../_example_code/usage_create_materials_factory.py
+:lines: 3-22
+:caption: Create material objects by setting a design code and calling factory functions.
+:::
+::::
+
+Notice how the factory functions accept the same keyword arguments as the constructors of the specific classes. This allows us to pass e.g. `gamma_c` and `alpha_cc` to `create_concrete`. Also notice how we are using {func}`reinforcement_duct_props() <structuralcodes.codes.ec2_2004.reinforcement_duct_props>` to get a dictionary with properties related to the specified ductility class of reinforcement.
+
+After creating a material object, it's properties are easily accessed, e.g. `concrete.Ecm`. Most notably, the `.constitutive_law` attribute reveals the [constitutive law](#usage-constitutive-laws) of the material.
```
talledodiego: typo? **its** properties?

docs/_example_code/usage_create_surface_geometries.py
```diff
@@ -0,0 +1,41 @@
+"""Example code for creating surface geometries."""
```
talledodiego, line 1: This is one possibility for creating a T geometry and is particularly suitable when for instance web and flage have different materials. When the material is the same it is possible to create a T section with shapely and create a single geometry with the same material. Perhaps it is woth mentioning this or adding an example showing the two approaches?

docs/usage/geometries/index.md
```diff
@@ -0,0 +1,140 @@
+(usage-geometries)=
+# Geometries
+
+## General
+
+Geometry objects are responsible for storing a polygon, a point, or a _compound_ of two or more of the former, and their respective material objects. Geometry objects are therefore the physical representation of a [section](#usage-sections).
+
+## Surface geometries
+
+A {class}`SurfaceGeometry <structuralcodes.geometry.SurfaceGeometry>` is initialized by passing a {class}`shapely.Polygon` with an arbitrary shape, and a [material object](#api-materials). See the example below where we create two rectangular surfaces and add them together to form a T-shaped geometry. Note that adding two or more {class}`SurfaceGeometry <structuralcodes.geometry.SurfaceGeometry>` objects results in a {class}`CompoundGeometry <structuralcodes.geometry.CompoundGeometry>` object.
```
talledodiego: See previous comment: it is worth mentioning that one can works before in shapely if only one material is needed?

docs/usage/geometries/index.md
```diff
@@ -0,0 +1,140 @@
...
+
+:::{tip}
+
+If you work in a Jupyter notebook, and return the geometry object in the final line of a code cell, the shape is visualized in the output below the code cell ✨
+:::
+
+:::{admonition} Creating holes
+:class: tip
+
+If you ever need to create a hole in a geometry, simply create a separate geometry object for the hole, and subtract it from the other. This operation returns a new {class}`CompoundGeometry <structuralcodes.geometry.CompoundGeometry>` with a hole ✅
```
talledodiego, line 46: This is a possibility, another possibility is including the whole in the polygon with the shapely syntax. Worth mentioning this?

docs/usage/geometries/index.md
```diff
@@ -0,0 +1,140 @@
...
+:::
+
+## Rectangular and circular geometries
+
+To simplify creating common geometrical shapes, we can use the {class}`RectangularGeometry <structuralcodes.geometry.RectangularGeometry>` or {class}`CircularGeometry <structuralcodes.geometry.CircularGeometry>` classes. These classes are wrappers for creating surface geometries with either a rectangular or a circular shape.
+
+Using {class}`RectangularGeometry <structuralcodes.geometry.RectangularGeometry>`, the above example can be simplified to the example below, where the changed lines are highlighted.
```
talledodiego, line 55: so cool highlighting the code lines!!!!!! 🤩

docs/usage/geometries/index.md
```diff
@@ -0,0 +1,140 @@
...
+::::
+
+## Point geometries and ways to add reinforcement
+
+A {class}`PointGeometry <structuralcodes.geometry.PointGeometry>` is a geometrical object defined by a coordinate, a diameter, and a material. Point geometries are used to represent reinforcement, and can be included in arbitrary locations, and in arbitrary numbers.
+
+There are several ways to create point geometries and add these to other geometry objects as reinforcement. Each of the following options results in a {class}`CompoundGeometry <structuralcodes.geometry.CompoundGeometry>` object holding the geometry object(s) that are reinforced, and the point geometries that act as reinforcement.
+
+- Create {class}`PointGeometry <structuralcodes.geometry.PointGeometry>` objects and add (`+`) them together with the geometry they reinforce.
+- Use the {func}`add_reinforcement() <structuralcodes.geometry.add_reinforcement>` function to add one {class}`PointGeometry <structuralcodes.geometry.PointGeometry>` to the geometry.
```
talledodiego, line 73: note for later:
now that I see this, it could be helpful have a general `add_reinforcements` methods where we can pass a list of coordinates to add several reinforcements in one?

docs/usage/sections/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+(code-usage-generic-section)=
+::::{dropdown-syntax}
+:::{literalinclude} ../../_example_code/usage_create_surface_geometries_with_reinforcement.py
+:lines: 12-13, 72-90
+:caption: Create a section and perform calculations.
+:::
+::::
+
+Notice how for example {py:meth}`.calculate_moment_curvature() <structuralcodes.sections.GenericSectionCalculator.calculate_moment_curvature>` returns a custom dataclass of type {class}`MomentCurvatureResults <structuralcodes.core._section_results.MomentCurvatureResults>`. If we inspect this class further, we find among its attributes `chi_y` and `m_y`. These are the curvature and the moment about the selected neutral axis. We also find the curvature and moment about the axis orthogonal to the neutral axis `chi_z` and `m_z`, and the axial strain at the level of the neutral axis `eps_axial`. All these attributes are stored as arrays, ready for visualization or further processing.
```
talledodiego: "These are the curvature and the moment about the selected neutral axis" this is not true I thing. This is respect to the global axis of the section right? In the sense that if we use `theta` different than zero, we will have `chi_y` that is not respect to the neutral axis, but respect to the y axis of the section in the CRS we defined. Does it make sense?

talledodiego (changes requested):
Excellent job! It is really inspiring to see this usage guide that shows how _structuralcodes_ can be used for tackling several problems!

For future (maybe we should create an issue) we should still:

- [ ] Cover how to compute cracked properties with linear behavior setting the desired "n" value
- [ ] Cover something about initStrain that we touched recently (pre-stressed section ultimate, pre-stressed initial phase with and without bonded cables).
- [ ] ...

docs/usage/geometries/index.md
mortenengen: I agree, this would be a useful feature. I created #267 and assigned it to you, if you do not mind? 😄

docs/usage/materials/index.md
mortenengen: Thanks for spotting this. Updated in 8b43d01.

docs/_example_code/usage_create_surface_geometries.py
mortenengen: Thanks for pointing out this. I have updated the example code in 8b43d01, and I think it flows better. I also added a new tip below the code.

When I wrote the draft, I already had the simplification of adding two RectangularGeometries in mind, and was kind of stuck with that modelling approach 😃

docs/usage/geometries/index.md
mortenengen: Good idea, I mentioned this in the tip below the code.

docs/usage/geometries/index.md
mortenengen: Absolutely. Thanks for pointing it out. I updated the tip in 8b43d01 😃

docs/usage/geometries/index.md
mortenengen: I always appreciate when others highlight stuff to indicate differences, and figured we should try to do the same 😄

docs/usage/sections/index.md
mortenengen: You are completely right, this was just poor wording from my side. Thanks for pointing it out 😃  Fixed in 8b43d01.

PR #284, opened 2025-09-17
feat: support arrays in the get_secant method on the base constitutive law

mortenengen:
- [x] Support array in `get_secant` method.
- [x] Add tests.

### DIFF
--- a/tests/test_materials/test_constitutive_laws.py
+++ b/tests/test_materials/test_constitutive_laws.py
@@ -638,7 +638,7 @@ def test_initial_strain_get_stress_tangent_no_strain_compatibility():
 def test_get_secant_in_base(eps):
     """Test the get_secant method in the base constitutive law."""
     # Arrange
-    constitutive_law = Elastic(E=30000)
+    constitutive_law = ParabolaRectangle(fc=45)
     if np.isscalar(eps):
         expected_secant = (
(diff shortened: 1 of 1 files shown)


tests/test_materials/test_constitutive_laws.py
```diff
@@ -629,3 +629,32 @@ def test_initial_strain_get_stress_tangent_no_strain_compatibility():
...
+
+
+@pytest.mark.parametrize(
+    'eps',
+    (-1e-3, 1.5e-3, 0, [-0.5e-3, 0.0e-3, 2.1e-3]),
+)
+def test_get_secant_in_base(eps):
+    """Test the get_secant method in the base constitutive law."""
+    # Arrange
+    constitutive_law = Elastic(E=30000)
```
talledodiego: may it be helpful using also a different constitutive law to test the non-trivial case when secant = tangent in all points?

talledodiego (approved):
Is it ok for me for merging! maybe just a small comment in the test to make it slightly more robust?

tests/test_materials/test_constitutive_laws.py
mortenengen: Updated. I changed the constitutive law to parabola rectangle and forced calling the method on the base class.

PR #286, opened 2025-09-29
feat: use warning and not exception to indicate no convergence

mortenengen:
This is an experiment to use warnings instead of exceptions to indicate that an iterative solution does not reach convergence.

When no convergence is reached, it can be useful to return some calculation results to the user, so that these results can be examined, and measures can be implemented.

In the current implementation, no convergence is indicated by raising an exception, and therefore, no results are returned from the iterative solver. In this PR, a custom warning is shown instead, and the results at the time of the warning was shown are returned to the user.

To make sure that the warning is always shown, a custom filter is added. The user can always override this behavior, e.g. only showing it once, completely ignoring it, or even raising it as an error.

### DIFF
--- a/structuralcodes/sections/_generic.py
+++ b/structuralcodes/sections/_generic.py
@@ -14,6 +14,7 @@ from shapely.ops import unary_union
 
 import structuralcodes.core._section_results as s_res
 from structuralcodes.core.base import Section, SectionCalculator
+from structuralcodes.core.errors import NoConvergenceWarning
 from structuralcodes.geometry import (
     CompoundGeometry,
     PointGeometry,
@@ -1474,6 +1475,12 @@ class GenericSectionCalculator(SectionCalculator):
                 break
 
         if num_iter >= max_iter:
-            raise StopIteration('Maximum number of iterations reached.')
+            warnings.warn(
+                message=(
+                    'Maximum number of iterations reached. '
+                    'Please review the strain profile carefully.'
+                ),
+                category=NoConvergenceWarning,
+            )
 
         return strain.tolist()
--- a/structuralcodes/__init__.py
+++ b/structuralcodes/__init__.py
@@ -1,7 +1,10 @@
 """A Python package that contains models from structural design codes."""
 
+import warnings
+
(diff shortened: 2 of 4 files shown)


talledodiego (changes requested):
The idea seems solid to me.
I tested it in a jupyter notebook and it seems to show correctly warnings each time.
We can probably extend this to other algorithms where we raise an exception for reaching the maximum number of iterations.

talledodiego:
> Would you like to explore the idea in other algorithms? You have more insight in the other methods on the `section_calculator` than me, so perhaps you also have some opinions on where we could replace exceptions with warnings?

Sure! I will work on this same PR. Hopefully during tonight's session

structuralcodes/sections/_generic.py
```diff
@@ -787,6 +955,92 @@ def calculate_bending_strength(
...
+        max_iter,
+        tol,
+    ):
+        # Find ultimate curvature from the strain distribution
+        # corresponding to failure and equilibrium with external axial
+        # force
+        # This coul not converge, let's catch the warning to customize
+        # the message
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('always', NoConvergenceWarning)
```
mortenengen, line 977: I think we should avoid overriding the filter inside a method. This way we might cause unexpected behavior for the user.

talledodiego: I am not sure I got this one correctly.

This do override user preference only within the `with`. This is necessary to be sure catching the warning independently from user preference. 
On exit of with the user preferences should be restored. So for instance if the user is treating warnings as once only he would not see (if it is not the first time it is raised), while if he/she is treating warnings as exception he/she will have an exception when I raise it (by the way in this scenario the user would remain with nothing to inspect on, and this is way I was suggesting to consider using exceptions encapsulating custom data for permitting inspection, so having a standard and uniform behavior).

mortenengen: Ok, got it. I misunderstood the context manager, but now I understand that it copies and upon exit restores the original warnings filter.

PR #289, opened 2025-10-04
Experiment/results-object

talledodiego:
This is just an experiment of what I was trying before our discussion.

### DIFF
--- a/structuralcodes/core/_section_results.py
+++ b/structuralcodes/core/_section_results.py
@@ -4,6 +4,7 @@ from __future__ import annotations  # To have clean hints of ArrayLike in docs
 
 from dataclasses import dataclass, field, fields
 
+import numpy as np
 from numpy.typing import ArrayLike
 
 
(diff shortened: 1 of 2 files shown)


structuralcodes/core/_section_results.py
```diff
@@ -160,8 +160,186 @@ class MomentCurvatureResults:
...
+            section=self.section,
+            eps_a=self.eps_axial[0],
+            chi_y=self.chi_y[0],
+            chi_z=self.chi_z[0],
+            n=self.n,
+            m_y=self.m_y[0],
+            m_z=self.m_z[0],
+            num_points=num_points,
+            seed=self.seed,
+        )
```
mortenengen: Could this portion of the code be refactored out to avoid the code repetition below? We could e.g. create an internal method `_create_detailed_result` which sets `self.detailed_result` based on `self.current_step`, so for each of the methods `create_detailed_result`, `next_step`, `previous_step` and `set_step` we can simply set `self.current_step` and then call the internal method.

structuralcodes/core/_section_results.py
```diff
@@ -160,8 +160,186 @@ class MomentCurvatureResults:
...
+            surface_data.append(
+                {
+                    'y': y,
+                    'z': z,
+                    'strain': strain,
+                    'stress': stress,
+                    'name': getattr(surf, 'name', None),
+                    'group_label': getattr(surf, 'group_label', None),
+                }
+            )
```
mortenengen, line 545: It would be great if we could store the surface data in a hierarchy similar to the point data, e.g. as a dict where the key is the group_label and the value is the dict with data.

Alternatively, and consistently for surface and point data, we could store the data in a manner that is easily input to a dataframe. For this structure, I guess one option could be to store a dict with the following keys: `x`, `y`, `strain`, `stress`, `group_label`, and each value is an array or a list of values for each individual point. So one dict for surface data and one for point data. This would open for really smooth filtering and plotting downstream.

tests/test_geometry/test_geometry.py
mortenengen, line 1: There are still lines that are not covered in testing. Could we perhaps add a test where we for example create detailed results for a moment curvature relation, and for each set of detailed results, integrate over the results to check if the detailed results add up to the moment, axial force, curvature and axial strain of the moment curvature relation?

structuralcodes/core/_section_results.py
talledodiego: I see your comment, thanks for it! I agree and tried to work in this direction.

talledodiego: Refactored on 86ddc75be8b0e5c783e047153bc7c70cae26243d

tests/test_geometry/test_geometry.py
talledodiego: Great idea. Will do as I finish implementing methods.

structuralcodes/core/_section_results.py
talledodiego: You are absolutely right! refactored in  5ec4322bbe711c4e337841d990a1c23008248d4d

structuralcodes/core/_section_results.py
```diff
@@ -145,6 +148,116 @@ def isclose(self, other, rtol=1e-5, atol=1e-8):
         return np.allclose(a, b, rtol=rtol, atol=atol)
 
 
+def _matching_geometries(
+    section,  # @mortenengen here we have circular import if I want to type it. Probably we should refactor? # noqa: E501
```
talledodiego: @mortenengen I have this doubt here. We should consider moving this code? Or do you have any other suggestion here?

tests/test_geometry/test_geometry.py
talledodiego: Actually it is not easy to create this test. Because we have a bunch of points, no area information, so we can't integrate... I implemented a different test actually that is not as nice as the one that you suggested. But it basically compares from some known points the results from get_point_stress and the detailed_result object.

docs/usage/sections/stress_field.png
mortenengen, line 1: Not really relevant for the implementation, but rather for the visualization in the docs. When using a diverging colormap like for the reinforcement stresses in this figure, we should make sure that the colormap is centered around zero. Otherwise, we should rather use a sequential colormap to get a better visualization of the intensity.

docs/usage/sections/strain_field.png
mortenengen, line 1: In this figure we could consider using a diverging colormap instead of a sequential one, see [matplotlib documentation](https://matplotlib.org/stable/users/explain/colors/colormaps.html). Remember to center the diverging colormap around zero 😃

docs/usage/sections/stress_field.png
talledodiego: done!

docs/usage/sections/strain_field.png
talledodiego: done!

PR #292, opened 2025-10-14
feat: Add parallel material and tests

talledodiego:
This PR contains an attempt to create another special wrapper for constitutive laws.

This wrapper models a set of constitutive laws as in parallel. This model assumes equal strains for all sub-materials and the total stress and tangent is additive. Optionaly weights can be provided.

An example is:

```
from structuralcodes.materials.constitutive_laws import ElasticPlastic, Parallel

import numpy as np

law1 = ElasticPlastic(E=1, fy=1, Eh=0, eps_su=2)
law2 = ElasticPlastic(E=0.5, fy=0.7, Eh=0.1, eps_su=3)
lawP = Parallel([law1, law2])

eps = np.linspace(0, 3.5, 100)
plt.plot(eps, law1.get_stress(eps), label='law1')
plt.plot(eps, law2.get_stress(eps), label='law_2')
plt.plot(eps, lawP.get_stress(eps), "k:", label='parallel')
plt.legend()
```

The result is shown here:
image

This opens the door to add easily tensile strength to concrete permitting to couple any model in tension (e.g. linear softening, exponential softening, etc.) with any model in compression avoiding the necessity of creating many versions of similar constitutive laws maybe differing only by the type of softening.

As an example: (keep in mind for now we don't have tensile laws that will be the topic of futre PRs so I am using a UserDefined law):

```
from structuralcodes.materials.constitutive_laws import ParabolaRectangle, UserDefined

law1 = UserDefined([-0.1, 0, 0.0002], [0, 0, 5])
law2 = ParabolaRectangle(-20)
lawP = Parallel([law1, law2])

eps = np.linspace(-0.005, 0.0003, 1000)
plt.plot(eps, law1.get_stress(eps), label='tension')
plt.plot(eps, law2.get_stress(eps), label='compression')
plt.plot(eps, lawP.get_stress(eps), "k:", label='parallel')
plt.legend()
```

Which shows the following expected result:
image

A very simple example using a section is given in the following lines:
```
from structuralcodes.materials.constitutive_laws import Parallel, ParabolaRectangle, UserDefined
from structuralcodes.materials.concrete import ConcreteMC2010

const_law_1 = ParabolaRectangle(-20)  # concrete in compression
const_law_2 = UserDefined([-0.01, 0, 0.0002], [0, 0, 5])  # concrete in tension
const_law = Parallel([const_law_1, const_law_2])

ConcreteMC2010(fck=30, constitutive_law=const_law)

geo=RectangularGeometry(width=100, height=100, material=mat)
section = GenericSection(geometry=geo)

res = section.section_calculator.integrate_strain_profile((0, 4e-5, 0))
```

The `get_stress` and `get_tangent` methods are trivial and the use of this constitutive law wrapper in a fiber approach is super easy. It was trickier making work the code for marin integrator because I had to write a function to "combine" all the marin coefficients and the strain ranges of all the materials. For each range, the terms of the polynomial functions for the same degree (i.e. exponent) are summed through all the constitutive laws.

As a final comment, this is a `ConstitutiveLaw`, not a `Material`. This constituve law can be created manually and assigned to any material (as the example above shows for `ConcreteMC2010`. I did not find the necessity of creating a `ParallelMaterial` (passing two materials) because  it becomes more slippery understading what a `ParallelMaterial` is and what are its properties. So this is only a `Constitutive` law that can be assigned to a material, not a `Material` itself.

If this makes sense, before merging I will add the proper documentation.

### DIFF
--- a/tests/test_materials/test_constitutive_laws.py
+++ b/tests/test_materials/test_constitutive_laws.py
@@ -12,6 +12,7 @@ from structuralcodes.materials.constitutive_laws import (
     ElasticPlastic,
     InitialStrain,
     ParabolaRectangle,
+    Parallel,
     Popovics,
     Sargin,
     UserDefined,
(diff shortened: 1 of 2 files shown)


structuralcodes/materials/constitutive_laws/_parallel.py
```diff
@@ -0,0 +1,234 @@
...
+import numpy as np
+from numpy.typing import ArrayLike
+
+from ...core.base import ConstitutiveLaw
+
+
+class Parallel(ConstitutiveLaw):
+    """Class for a Parallel Constitutive Law."""
+
+    _strain_compatibility: bool = True
```
mortenengen: Is it necessary to set the internal attribute `_strain_compatibility`?

mortenengen (changes requested):
Good job on this really useful feature. I have added one minor comment to the code, and merged dev to resolve conflicts and added an entry in the API reference.

Could you please add more tests to cover all the added code lines?

structuralcodes/materials/constitutive_laws/_parallel.py
talledodiego: no you are right! removed!

PR #293, opened 2025-10-21
docs: update figures in theory docs

mortenengen:


docs/_source_fig/theory_figs.svg
talledodiego, line 1: should we put this file inside a folder like for instance _source_fig or something similar?

talledodiego (approved):
Excellent job! The figures are outstanding!!! Only one minor comment.

PR #294, opened 2025-10-21
docs: remove sections from theory with duplicate contents

mortenengen:


talledodiego (approved):
Thanks for doing this! Sounds perfect to me

PR #308, opened 2025-12-19
feat: add option to calculate the complete mn domain

mortenengen:
## Description

Add an option to calculate the complete MN interaction domain for a `GenericSection`.

## To do
- [x] Add test.
- [x] Add entry in usage guide.

### DIFF
--- a/structuralcodes/sections/_generic.py
+++ b/structuralcodes/sections/_generic.py
@@ -1093,9 +1093,10 @@ class GenericSectionCalculator(SectionCalculator):
                     type_6=type_6,
                 )
             )
-            # Here, we should make sure that
-            strains = np.array([*strains, *additional_strains[-2:0:-1]])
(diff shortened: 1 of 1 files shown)


structuralcodes/sections/_generic.py
```diff
@@ -1070,6 +1074,29 @@ def calculate_nm_interaction_domain(
...
+                    type_1=type_1,
+                    type_2=type_2,
+                    type_3=type_3,
+                    type_4=type_4,
+                    type_5=type_5,
+                    type_6=type_6,
+                )
+            )
+            # Here, we should make sure that
+            strains = np.array([*strains, *additional_strains[-2:0:-1]])
```
talledodiego: Strains and additional_strains are already numpy arrays. Is it better to concatenate them directly with numpy methods?

mortenengen: Good point, you are totally right! Fixed in c4f087e 😊

PR #309, opened 2026-01-07
feat: add Eq. M1 according to EUROCODE 1992-3:2006

reigt:


structuralcodes/codes/ec2_2004/_cracking_restraint_imposed_deformations.py
```diff
@@ -0,0 +1,37 @@
+"""Collection of functions from EUROCODE 1992-3:2006."""
+
+
+def eps_sm_eps_cm_m1(
```
mortenengen: Can you please consider renaming the function to e.g. "eps_sm_eps_cm_restraint_end"?

PR #311, opened 2026-01-20
feat: add time development of concrete properties according to EC2 2004

mortenengen:


structuralcodes/codes/ec2_2004/_concrete_material_properties.py
```diff
@@ -237,3 +242,65 @@ def fcd(fck: float, alpha_cc: float, gamma_c: float) -> float:
...
+    """The time development function for compressive strength of concrete.
+
+    EN 1992-1-1:2004, Eq. (3.2)
+
+    Args:
+        t (ArrayLike): The time in days to evaluate the development function
+            for.
+        s (float): The scale factor in the exponent for the time development
+            function. s = 0.20 for class R, 0.25 for class N, and 0.38 for
+            class N.
```
talledodiego, line 264: Alternatively one could have given a literal indicating the CEM and assign s with a dictionary. Anyhow I am totally fine with this implementation. 

For uniformity later we should check in other parts of the code because I think for similar cases we have the literal CEM for instance as the argument and s is computed with a dictionary.

structuralcodes/codes/ec2_2004/_concrete_material_properties.py
```diff
@@ -237,3 +242,65 @@ def fcd(fck: float, alpha_cc: float, gamma_c: float) -> float:
...
+
+    Returns:
+        ArrayLike: The value of the time development function.
+    """
+    return np.exp(s * (1 - np.sqrt(28 / t)))
+
```
talledodiego, line 270: Should we also include function (3.1) (i.e. fcm(t) = beta_cc(t) * fcm)? Or is it too simple to deserve an actual implementation?

structuralcodes/codes/ec2_2004/_concrete_material_properties.py
```diff
@@ -237,3 +242,65 @@ def fcd(fck: float, alpha_cc: float, gamma_c: float) -> float:
...
+    if np.isscalar(t):
+        beta = beta_cc(t, s)
+        if t < 28:
+            return beta
+        return np.pow(beta, 2 / 3)
+    t = np.atleast_1d(t)
+    beta = beta_cc(t, s)
+    beta[t >= 28] = np.pow(beta[t >= 28], 2 / 3)
+    return beta
+
```
talledodiego, line 296: Same comment as before. Should we implement the whole (3.4) ?

Moreover a slight consideration: we tend to map relations (or part of relations) 1:1 with the code. The argument in 3.4 is beta_cc? s is not strictly an argument of this function? I am not sure on this myself...

structuralcodes/codes/ec2_2004/_concrete_material_properties.py
```diff
@@ -237,3 +242,65 @@ def fcd(fck: float, alpha_cc: float, gamma_c: float) -> float:
...
+        t (ArrayLike): The time in days to evaluate the development function
+            for.
+        s (float): The scale factor in the exponent for the time development
+            function. s = 0.20 for class R, 0.25 for class N, and 0.38 for
+            class N.
+
+    Returns:
+        ArrayLike: The value of the time development function.
+    """
+    return np.pow(beta_cc(t, s), 0.3)
```
talledodiego, line 313: Same comment as before... The user would expect to get also Ecm(t)?

structuralcodes/codes/ec2_2004/_concrete_material_properties.py
mortenengen: Good point. I added a function for getting the s-value in d3a6d74.

structuralcodes/codes/ec2_2004/_concrete_material_properties.py
mortenengen: Added in f224056.

structuralcodes/codes/ec2_2004/_concrete_material_properties.py
mortenengen: I added the one-to-one mapping in f224056, but kept the implemented time development function for convenience and left a note in the docstring.

structuralcodes/codes/ec2_2004/_concrete_material_properties.py
mortenengen: I added the one-to-one mapping in f224056, but kept the implemented time development function for convenience and left a note in the docstring.

talledodiego (approved):
Excellent. I totally agree with your comments.

PR #314, opened 2026-02-04
refactor: use point.buffer to create circular polygon

andrefmello91:
Use shapely default way for creating a circle.

### DIFF
--- a/structuralcodes/geometry/_circular.py
+++ b/structuralcodes/geometry/_circular.py
@@ -22,7 +22,10 @@ def _create_circle(
     """Create a circle with a given radius."""
     origin = origin if origin is not None else (0.0, 0.0)
     pt = Point(origin)
-    quad_segs = 4 * round(npoints / 4)  # Round to the nearest multiple of 4
+    # Note that .buffer takes the argument quad_segs which is the number of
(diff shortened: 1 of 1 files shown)


structuralcodes/geometry/_circular.py
```diff
@@ -8,23 +8,19 @@
...
 from numpy.typing import ArrayLike
-from shapely import Polygon
+from shapely import Point
 
 from structuralcodes.core.base import Material
 
 from ._geometry import SurfaceGeometry
 
 
-def _create_circle(radius, npoints=20, origin: t.Optional[ArrayLike] = None):
```
mortenengen, line 20: Great if we could maintain the interface where we let the user specify the number of points along the circle.

I have noticed that the `.buffer` method accepts the optional argument `quad_segs` which controls the number of line segments in one quarter of the circle. Please consider exposing the `n_points` argument again, noting that it should be a multiple of 4, and if not, that we will round up to the nearest multiple of 4, and furthermore use this as basis for the argument `quad_segs` in `.buffer`.

andrefmello91:
Sorry for making the PR in the main branch.
I checked the error, but at the moment I could not figure what is causing it.

Issue #303, opened 2025-12-11
Bug in section.gross_properties in relation to calculate_moment_curvature()


elliotmelcer:
## calculate_moment_curvature() affects section.gross_properties
I have noticed that the gross section properties of a section channge if calculate_moment_curvature() is run before evaluating them. Here is my code and a visualization of the cross-section (not included in the provided code). This seems similar to issue #297. 

```from shapely import Polygon
from structuralcodes import set_design_code
from structuralcodes.geometry import SurfaceGeometry, add_reinforcement_line
from structuralcodes.materials.concrete import create_concrete
from structuralcodes.materials.reinforcement import create_reinforcement
from structuralcodes.sections import GenericSection

set_design_code('ec2_2004')

# hp_geometry stuff
b  = 300      #mm
b1 = 50
b0 = 200
d = 200
d1 = 200
d0 = 400
cover = 50

polygon = Polygon(
    [
        (-b /2,  d/2),
        (-b /2, -d/2),
        (-b0/2, -d/2),
        (-b0/2, -d/2-d1),
        ( b0/2, -d/2-d1),
        ( b0/2, -d/2),
        ( b /2, -d/2),
        ( b /2,  d/2),
    ]
)

# --- Material Properties ---
# concrete
concrete_c40= create_concrete(fck=40, constitutive_law = 'parabolarectangle', alpha_cc = 0.85, gamma_c = 1.5)

# steel reinforcemenet
reinforcement = create_reinforcement(fyk=500, Es=200000, ftk=500, epsuk=0.025, density=7850, constitutive_law="elasticperfectlyplastic")

# --- concrete section ---
geometry = SurfaceGeometry(
    poly=polygon, material=concrete_c40
)

geometry = add_reinforcement_line(
    geometry,
    (-b0/2+cover, -d/2 - d1 + cover),
    (b0/2-cover, -d/2 - d1 + cover),
    20,
    reinforcement,
    3
)

def return_section_properties(sec: GenericSection) -> str:
    # --- Geometric Properties ---
    gross_props = sec.gross_properties

    cz = gross_props.cz
    iyy = gross_props.iyy
    e_iyy = gross_props.e_iyy
    ea = gross_props.ea

    props = (f"I_yy={iyy:.2e} mm^4, "
          f"EA={ea:.3e} N, "
          f"EI_yy={e_iyy:.4e} Nmm^2, "
          f"cz={cz:.2e} mm, ")
    return props


[...]

PR #315, opened 2026-02-10
fix: Gross properties when computing after a section calculation

talledodiego:
This solves issue #303 

The wrong `integration_data` was passed during the computation of gross properties.

Included a test showing the solution of the issue.


mortenengen (approved):
Thanks for solving this issue!

PR #321, opened 2026-03-04
feat: filtering geometries based on name and group label

talledodiego:
Add filtering potentialities based on name or group label used for geometries.

This supports passing `None` as a pattern string and get back all geometries. The `return_mode` can be `"split"` or `"flat"`. The former will return a dictionary with surface and point geometries separated, the second will return a single flattened list.

By default the match is case sensitive, but it can be optionally set to case insensitive if needed by the user.

Wildcards lie `'*'` or `'?'` are accepted. The backend used for matching is fnmatch.

Some examples are:

`geo.name_filter(None)` This will return the list of all surface and point geometries in the compound `geo`.
`geo.name_filter("we*")` This will return the list of all surfaces and point geometries which name matches the pattern `"we*"`
`geo.name_filter("web", return_mode="split")` This will return a dictionary with keys `'surfaces'` and `'points'` with the lists of surfaces and points geometries matching the pattern `"web"`
`geo.name_filter("*geo*", case_sensitive=False)` This will return the list of all surfaces and point geometries which name mathes the pattern `"*geo*"` case insensitive.
`geo.group_filter("*reinf*")` This will return the list of all surfaces and point geometries which group_name matches the pattern `"*reinf*"`

It is possible also to combine filters. For instance we can get the list of geometries matching different patterns for name and group_label:
```
geom_name = geo.name_filter("*geo*", case_sensitive=False)
geom_group = geo.group_filter("bottom*")

list(set(geom_name) & set(geom_group))
```

The implementation requires the following steps:

- [x] Implement the code for the feature at the geometry level
- [x] Add proper tests
- [x] Include the documentation

### DIFF
--- a/structuralcodes/geometry/_geometry.py
+++ b/structuralcodes/geometry/_geometry.py
@@ -4,6 +4,7 @@ from __future__ import annotations  # To have clean hints of ArrayLike in docs
 
 import typing as t
 import warnings
+from fnmatch import fnmatchcase
 from math import atan2
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


structuralcodes/geometry/_geometry.py
```diff
@@ -933,3 +996,148 @@ def from_geometry(
...
+                if g._group_matches(pattern, case_sensitive=case_sensitive)
+            ]
+            points = [
+                g
+                for g in points
+                if g._group_matches(pattern, case_sensitive=case_sensitive)
+            ]
+        if return_mode == 'flat':
+            return [*surfaces, *points]
+        return {'surfaces': surfaces, 'points': points}
```
mortenengen, line 1135: We seem to have left out testing group filtering with split returns. Please consider adding a test that covers this.

mortenengen (changes requested):
Thanks for the nice contribution! I have left only one minor comment that you could consider while finalizing.

structuralcodes/geometry/_geometry.py
talledodiego: Right, done in 62f9e47b43d5d2892154834db06b44e552599c93

talledodiego:
Thanks for the review! included two tests covering the case you evidenced.

mortenengen (approved):
Very good 😃

PR #324, opened 2026-03-12
chore: add html export of coverange in makefile

talledodiego:
With this PR I am proposing to add the html report of coverage.


mortenengen (approved):
Thanks for adding this!

PR #325, opened 2026-03-13
feat: detailed results for calculate strain profile

talledodiego:
With this PR I implemented custom results for `calculate_strain_profile`.

The objects store information (some are calculated only if and when requested) to assess the convergence process, e.g. the following lines:

```
plt.semilogy(strain_res.residual_norm_history, "-o")
plt.ylabel(r"$||\mathbf{r}||$")
```

plot the norm of the residual vector:

image

A future PR will assess the results for `integrate_strain_profile`


mortenengen (approved):
Thanks for this useful contribution!

PR #326, opened 2026-03-13
feat: add custom result objects to integrate_strain_profile

talledodiego:
This is a proposal for adding a custom result object  also for method integrate_strain_profile. In this case two different results are available: `IntegrateStrainStiffnessResult` when integrating modulus, and `IntegrateStrainForceResult` when integrating stresses.
They both have methods like `asarray` for having the array representation.

Optionally we could evaluate if needed to implement

```
def __iter__(self):
    yield self.n
    yield self.m_y
    yield self.m_z

def __len__(self):
    return 3

def __getitem__(self, i):
    return (self.n, self.m_y, self.m_z)[i]
```

But I have the feeling it might be less transparent.

### DIFF
--- a/structuralcodes/core/_section_results.py
+++ b/structuralcodes/core/_section_results.py
@@ -7,7 +7,7 @@ import typing as t
 from dataclasses import dataclass, field, fields
 
 import numpy as np
-from numpy.typing import ArrayLike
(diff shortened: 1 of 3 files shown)


mortenengen (approved):
Thanks for this useful contribution!

PR #328, opened 2026-03-17
feat: improve result objects

talledodiego:
@mortenengen, as requeste:
- included docs
- increased test coverage
- refactored interaction domain


mortenengen (approved):
Thanks for taking the time to do these improvements!

PR #331, opened 2026-03-17
chore: refactor class hierarchy to better avoid circular imports

mortenengen:
This is an attempt to refactor the class hierarchy so that future additions to the library are less likely to cause circular imports.


talledodiego (approved):
Thanks for this. I totally agree with the approach of splitting the different base classes.

PR #337, opened 2026-03-19
bugfix: add return to chi_y property in IteractionDomainResult

talledodiego:
A return was missing


mortenengen (approved):
Thanks for spotting and fixing this!

Issue #329, opened 2026-03-17
AttributeError: 'MultiPolygon' object has no attribute 'exterior' in gross_properties


MestreCarlos:
Hi, I found some little bug when computing `gross_properties` for a section whose geometry is a `MultiPolygon` (e.g. disconnected rectangles), an `AttributeError` is raised because `MultiPolygon` has no `exterior` attribute. It’s a minor thing and I don’t think it’s necessary to create a new PR for this. 

## Location
`structuralcodes/sections/_generic.py`, method `_calculate_gross_section_properties`,

## Problematic code

```
if isinstance(polygon, MultiPolygon):
    gp.perimeter = 0.0
    warnings.warn(
        'Perimiter computation for a multi polygon is not defined.'
    )

gp.perimeter = polygon.exterior.length  # ← Always executed; fails with MultiPolygon
```

The if block sets gp.perimeter = 0.0 for MultiPolygon, but the next line always runs and accesses polygon.exterior, which does not exist on MultiPolygon.

## Proposed fix
```
if isinstance(polygon, MultiPolygon):
    gp.perimeter = 0.0
    warnings.warn(
        'Perimiter computation for a multi polygon is not defined.'
    )
else:
    gp.perimeter = polygon.exterior.length
```

## Code example
```
from shapely import Polygon

from structuralcodes.geometry import CompoundGeometry, SurfaceGeometry
from structuralcodes.sections import GenericSection
from structuralcodes.materials.basic import ElasticMaterial


mat = ElasticMaterial(E=210000, density=7850)
rect1 = SurfaceGeometry(
    poly=Polygon([(0, 0), (100, 0), (100, 10), (0, 10)]),
    material=mat,
)
rect2 = SurfaceGeometry(
    poly=Polygon([(0, 200), (100, 200), (100, 210), (0, 210)]),
    material=mat,
)

compound = CompoundGeometry([rect1, rect2])
section = GenericSection(compound, integrator="marin")
gp = section.gross_properties

talledodiego:
Thanks @MestreCarlos for pointing this out! indeed I forgot the else...

Fixed in #332

talledodiego:
@mortenengen I think this can be closed now that #332 has been merged with dev.

PR #332, opened 2026-03-18
fix: bugfix perimeter

talledodiego:
This PR solves the bug pointed out in #329.

PR #338, opened 2026-03-22
fix: guard perimeter computation for MultiPolygon geometries

saschabuehrle:
## Bug
[#329](https://github.com/fib-international/structuralcodes/issues/329) — `gross_properties` raises `AttributeError: 'MultiPolygon' object has no attribute 'exterior'` when the section geometry is a `MultiPolygon` (e.g. disconnected rectangles).

## Fix
The `if isinstance(polygon, MultiPolygon)` block correctly sets `gp.perimeter = 0.0` and emits a warning, but the code then falls through to `polygon.exterior.length` unconditionally. This adds an `else` branch so the `.exterior` access only runs for single `Polygon` geometries.

## Testing
Verified that the code path now correctly short-circuits for `MultiPolygon` and only accesses `.exterior` on single `Polygon` instances. The existing test suite continues to pass for single-polygon sections.

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/structuralcodes/sections/_generic.py
+++ b/structuralcodes/sections/_generic.py
@@ -150,8 +150,8 @@ class GenericSectionCalculator(SectionCalculator):
             warnings.warn(
                 'Perimiter computation for a multi polygon is not defined.'
             )
-
-        gp.perimeter = polygon.exterior.length
+        else:
+            gp.perimeter = polygon.exterior.length
 
         # Computation of area: this is taken directly from shapely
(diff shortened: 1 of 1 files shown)


talledodiego:
Thanks @saschabuehrle for taking the time to fix this.
We fixed this in PR #332 already merged with dev.

Thanks for your time!

saschabuehrle:
Ah, missed that #332 already landed. Closing — thanks for letting me know!

PR #340, opened 2026-03-25
fix: update iterative strain before convergence check

mortenengen:
After implementing the result object for calculate strain plane, the order of checking for convergence and updating the iterative strain was changed, resulting in a non-converged strain plane to be returned from the method.

### DIFF
--- a/structuralcodes/sections/_generic.py
+++ b/structuralcodes/sections/_generic.py
@@ -1829,13 +1829,13 @@ class GenericSectionCalculator(SectionCalculator):
                 # Solve using the current tangent stiffness
                 delta_strain = np.linalg.solve(stiffness_tangent, residual)
 
+            # Update the strain
+            strain += delta_strain
+
             # Check for convergence:
             if np.linalg.norm(delta_strain) < tol:
                 converged = True
                 break
-
-            # Update the strain
-            strain += delta_strain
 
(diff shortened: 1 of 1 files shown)


structuralcodes/sections/_generic.py
```diff
@@ -1790,9 +1790,14 @@ def calculate_strain_profile(
...
         # Calculate strain plane with Newton Rhapson Iterative method
         num_iter = 0
-        strain = np.zeros(3, dtype=float)
-        residual = np.zeros(3, dtype=float)
         converged = False
+        strain = np.zeros(3, dtype=float)
+
+        # Calculate the initial response and residuals. Note that the initial
+        # residual might be different from the applied loads if any initial
+        # strain is present
```
talledodiego, line 1798: right observation! perfect

talledodiego (approved):
Perfect! Thanks for having catched this up!
Sorry for having not catched this before hand.

Should we in a next PR make deeper tests on this?

mortenengen:
> Perfect! Thanks for having catched this up!
> Sorry for having not catched this before hand.
> 
> Should we in a next PR make deeper tests on this?

Thanks😊 It would be great if we could have a couple of tests that check the actual valeus of the strain plane.

PR #343, opened 2026-04-08
feat: add ACI 318-19 concrete and reinforcement material properties

### LINKED ISSUE #187: Plans for future code adoption?
Great work on this project.

Are there any plans for incorporating other international codes? US, Canada, Australia?

gabe-kafka:
## Summary

Adds the first American design code to the library: ACI 318-19 material properties from Chapters 19, 20, and 22.

- **Design code module**: `structuralcodes.codes.aci318_19`, registered as `aci318_19`.
- **Concrete functions**: `Ec`, `fr`, `lambda_factor`, `eps_cu`, `eps_c0`, `alpha1`, and `beta1` as standalone functions following existing EC2/MC2010 patterns. `fct` is intentionally not included because ACI 318-19 does not define it.
- **Reinforcement functions**: `Es`, `fy_design`, `epsyd`, and `reinforcement_grade_props` for the ACI 318-19 Table 20.2.1.3(a) grade lookup.
- **Material classes**: `ConcreteACI318_19` and `ReinforcementACI318_19` with constitutive law support (elastic, parabola-rectangle, and bilinear for concrete; elastic and elastic-plastic for steel).
- **ACI customary inputs**: unit conversion helpers plus `ConcreteACI318_19.from_psi(...)`, `ReinforcementACI318_19.from_grade(...)`, and `ReinforcementACI318_19.from_ksi(...)`; values are converted at the API boundary and stored internally in SI units.

All direct code/material APIs use SI internally (MPa, kg/m3). `ConcreteACI318_19` accepts ACI `fc` directly; `fck` remains available only for compatibility with the shared concrete factory/base API.

## ACI vs Eurocode safety philosophy

ACI 318 uses LRFD. Strength reduction factors (`phi`) are applied at member capacity level, not material level. The ACI material classes keep `gamma_c` and `gamma_s` for compatibility with the shared material APIs, but values other than `None` or `1.0` now raise `ValueError`.

## Scope

Material properties only. No member design (shear, flexure, etc.) in this PR; that can follow once the architecture question in #187 is resolved.

Closes #187 (partially - first code added, more can follow).

## Test plan

- [x] `./venv/bin/python -m ruff check structuralcodes/codes/aci318_19 structuralcodes/materials/concrete/_concreteACI318_19.py structuralcodes/materials/reinforcement/_reinforcementACI318_19.py tests/test_aci318_19` - passes
- [x] `./venv/bin/python -m ruff format --check structuralcodes/codes/aci318_19 structuralcodes/materials/concrete/_concreteACI318_19.py structuralcodes/materials/reinforcement/_reinforcementACI318_19.py tests/test_aci318_19` - passes, 11 files already formatted
- [x] `./venv/bin/python -m pytest tests/test_aci318_19` - 84 passed
- [x] `./venv/bin/python -m pytest` - ran locally; 10,240 passed and 1 EC2 2023 expectation mismatch reproduced only in this workspace so far, not in CI: `tests/test_ec2_2023/test_ec2_2023_section5_materials.py::test_eps_cs_50y[25-dry-800-CN-0.46333]` returned `0.45333333333333337` vs expected `0.46333`
- [x] Round-trip: `set_design_code('aci318_19')` -> `create_concrete(fck=28)` -> `.Ec` returns the ACI normalweight value
- [x] Factory dispatch works for both concrete and reinforcement
- [x] Concrete and reinforcement constitutive laws instantiate correctly


structuralcodes/codes/__init__.py
```diff
@@ -3,9 +3,10 @@
 import types
 import typing as t
 
-from . import ec2_2004, ec2_2023, mc2010, mc2020
+from . import aci318, ec2_2004, ec2_2023, mc2010, mc2020
```
mortenengen, line 6: Minor comment about the naming.

For Eurocode 2 we decided to implement different generations in different modules, e.g. `ec2_2004` for the first generation that was released in 2004, and `ec2_2023` for the second generation that was released in 2023. The second generation is expected to become accepted as national governing design code around Europe some time in the period 2027-2028.

Please correct me if I am wrong here, but my impression is that the 19 in ACI318-19 indicates that it was released in 2019. Similarly, a slightly newer version was released in 2025 and named ACI318-25. If this is right, please consider indicating this in the name of the module, e.g. `aci318_19` instead of just `aci318`.

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+
+import typing as t
+
+from structuralcodes.codes import aci318
+
+from ..constitutive_laws import ConstitutiveLaw, create_constitutive_law
+from ._concrete import Concrete
+
+
+class ConcreteACI318(Concrete):  # noqa: N801
```
mortenengen, line 11: Same minor comment as above, please consider renaming the class to `ConcreteACI318_19`.

structuralcodes/materials/reinforcement/_reinforcementACI318.py
```diff
@@ -0,0 +1,164 @@
...
+
+import typing as t
+
+from structuralcodes.codes import aci318
+
+from ..constitutive_laws import ConstitutiveLaw, create_constitutive_law
+from ._reinforcement import Reinforcement
+
+
+class ReinforcementACI318(Reinforcement):  # noqa: N801
```
mortenengen, line 11: Same minor comment as above, please consider renaming the class to `ReinforcementACI318_19`.

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+            initial_strain=initial_strain,
+            initial_stress=initial_stress,
+            strain_compatibility=strain_compatibility,
+        )
+        self._Ec = abs(Ec) if Ec is not None else None
+        self._fr = abs(fr) if fr is not None else None
+        self._fct = abs(fct) if fct is not None else None
+        self._wc = wc
+        self._lambda_s = lambda_s
+
```
mortenengen, line 134: In the other concrete classes we have implemented a `.__post_init__` method which is responsible for validating the parameters that are set. See for example [here](https://github.com/fib-international/structuralcodes/blob/1fed7cef5730af3a024c90d38a1ef0d6905cde3e/structuralcodes/materials/concrete/_concreteEC2_2004.py#L166). Please consider implementing a similar method here.

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+    def __parabolarectangle__(self) -> dict:
+        """Returns kwargs for a parabola-rectangle constitutive law.
+
+        Note:
+            Uses ACI 318 parameters: peak strain at 0.002,
+            ultimate strain at 0.003, parabolic exponent n=2.
+        """
+        return {
+            'fc': self.fcd(),
+            'eps_0': 0.002,
```
mortenengen, line 268: Is there a clause in the code that sets the value `eps_0 = 0.002`? If it is, please consider implementing a function in the code module similar as for `eps_cu`. Furthermore, the value of `eps_0` should preferably be taken from a property on this concrete class such that it is more transparent for the user what is routed to the constitutive laws. Please consider implementing this property, possibly calling the function for `eps_0` in the code module.

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+            the parabola-rectangle default. This is not the ACI
+            318 Whitney rectangular stress block. The library
+            integrates the parabola-rectangle curve directly,
+            which ACI 318-19 Section 22.2.2.3 permits as an
+            alternative to the Section 22.2.2.4 equivalent
+            rectangular stress block simplification.
+        """
+        return {
+            'fc': self.fcd(),
+            'eps_c': 0.002,
```
mortenengen, line 287: Same as for the comment related to `eps_0` above.

structuralcodes/codes/aci318/_reinforcement_material_properties.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Raises:
+        ValueError: If fy is not positive.
+    """
+    if fy <= 0:
+        raise ValueError(f'fy={fy} must be positive')
+    return fy / _Es
+
+
+def reinforcement_grade_props(
```
mortenengen, line 76: Does ACI 318-19 also provide values for the strain at ultimate strength? If so, please consider including it in the return dictionary.

mortenengen (changes requested):
Thanks for this very nice contribution @gabe-kafka. I have left a couple of comments for you to consider while finalizing the PR. Please note that @talledodiego and @aperezcaldentey have also agreed to provide their reviews.

mortenengen:
@gabe-kafka, in the above message you provided the following comment:

> 
>     * [x]  `make test` — 10,186 passed (73 new + 10,113 existing), 1 pre-existing failure in EC2 2023
> 

However, I have not experienced any pre-existing failues in the pipeline. Please elaborate so that we can triage this.

structuralcodes/codes/aci318_19/_concrete_material_properties.py
```diff
@@ -0,0 +1,164 @@
...
+        raise ValueError(
+            f'Unknown concrete type: {concrete_type}. '
+            f'Valid types: {list(LAMBDA_FACTORS.keys())}'
+        )
+    return result
+
+
+def fct(fc: float, lambda_s: float = 1.0) -> float:
+    """The approximate splitting tensile strength of concrete.
+
```
aperezcaldentey, line 138: The formula used in this method does not correspond to ACI 318-19, Section 19.2.4.3, as stated. We should possibly delete this method as fct is not defined in ACI 318-19.

structuralcodes/codes/aci318_19/_concrete_material_properties.py
```diff
@@ -0,0 +1,164 @@
+"""Concrete material properties according to ACI 318-19."""
+
```
aperezcaldentey, line 2: I suggest we order the methods as they appear in the code. Let's not jump from Chaper 19 to 22 and then back to 19.

structuralcodes/codes/aci318_19/_concrete_material_properties.py
```diff
@@ -0,0 +1,164 @@
...
+        ValueError: If fc is not positive.
+    """
+    if fc <= 0:
+        raise ValueError(f'fc={fc} must be positive')
+    if fc <= 28:
+        return 0.85
+    if fc >= 55:
+        return 0.65
+    return 0.85 - 0.05 * (fc - 28) / 7
+
```
aperezcaldentey, line 98: Replace expression: 0.85 - 0.05 * (fc - 28) / 7 by 0.85-0.20/27*(fc-28) - why simplify and lose accuracy? I could drive some people crazy...

structuralcodes/codes/aci318_19/_concrete_material_properties.py
```diff
@@ -0,0 +1,164 @@
...
+    Args:
+        fc (float): The specified compressive strength of concrete in
+            MPa.
+
+    Returns:
+        float: The stress block depth factor (dimensionless).
+
+    Raises:
+        ValueError: If fc is not positive.
+    """
```
aperezcaldentey, line 90: According to Table 22.2.2.4.3. f'c must be greater that 17. We should set this as a lower limit. Not 0.

structuralcodes/codes/aci318_19/_concrete_material_properties.py
```diff
@@ -0,0 +1,164 @@
...
+LAMBDA_FACTORS = {
+    'normalweight': 1.0,
+    'sand-lightweight': 0.85,
+    'all-lightweight': 0.75,
+}
+
+
+def Ec(fc: float, wc: float = 2320.0) -> float:
+    """The modulus of elasticity of concrete.
+
```
aperezcaldentey, line 17: Default value of concrete. The typical specific weight of concrete used in North American practice is 150 lbs/cf, which I roughly equivalent to a density of 150/(2.2*0.3048^3)≈2400 kg/m3. I understand this my have to do with going from specific weight to density using g=9.81 m/s2, but it is not clear for me how you derive it, as the units are not consistent. On the other hand, Eq. 19.2.2.1b provides a factor of 4700 which would lead to a value of w=2286 kg/m3.

structuralcodes/codes/aci318_19/_concrete_material_properties.py
```diff
@@ -0,0 +1,164 @@
...
+import math
+import typing as t
+
+LAMBDA_FACTORS = {
+    'normalweight': 1.0,
+    'sand-lightweight': 0.85,
+    'all-lightweight': 0.75,
+}
+
```
aperezcaldentey, line 9: The lambda factor in ACI-318-19 is a function of the weight and lies between 1.00 and 0.75. You provide 'sand-lightweight as an intermediate value. Consider programing the equation as a function of the weight (as done for the modulus of elasticity) or adding the values from Table 19.2.4.1 'lightweight, fine blend' and 'sand-lightweight, corase blend'

structuralcodes/codes/aci318/__init__.py
```diff
@@ -0,0 +1,37 @@
+"""ACI 318-19."""
+
+import typing as t
+
+from ._concrete_material_properties import (
+    Ec,
+    alpha1,
+    beta1,
+    eps_cu,
+    fct,
```
aperezcaldentey: fct not defined in ACI 318-19

structuralcodes/codes/aci318_19/_reinforcement_material_properties.py
```diff
@@ -0,0 +1,100 @@
...
+
+import typing as t
+
+REINFORCEMENT_GRADES = {
+    '40': {'fy': 280.0, 'fu': 420.0},
+    '60': {'fy': 420.0, 'fu': 550.0},
+    '80': {'fy': 550.0, 'fu': 690.0},
+    '100': {'fy': 690.0, 'fu': 860.0},
+}
+
```
aperezcaldentey, line 15: Values for Reinforcement grades based on table 20.2.1.3(a). Please add this reference to comments.

structuralcodes/codes/aci318_19/_reinforcement_material_properties.py
```diff
@@ -0,0 +1,100 @@
...
+    Returns:
+        float: The modulus of elasticity in MPa.
+    """
+    return 200000.0
+
+
+def fy_design(fy: float, phi: float = 1.0) -> float:
+    """The design yield strength of reinforcement.
+
```
aperezcaldentey, line 30: As you say, ACI does not consider a partial factor on steel. Why do we need to provide one? The user should not be able to introduce a value different that 1.00. If we need this method for compatibility reasons we shoudl delete phi as a KWARG.

structuralcodes/codes/__init__.py
aperezcaldentey: Morten, I agree with your comment. We need to keep different versions separate. When there is no change we can use the method of the previous version, but we may need to update the references and son on. So, yes it should be aci318-19

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+
+class ConcreteACI318(Concrete):  # noqa: N801
+    """Concrete implementation for ACI 318-19.
+
+    Note:
+        ACI 318 uses specified compressive strength (fc') rather than
+        characteristic strength (fck). The parameter is named fck for
+        compatibility with the base class, but represents fc' in ACI
+        notation. An fc property is provided as an alias.
+
```
aperezcaldentey: Note that fck is a 95% fractile, while f'c is a 90% fractile

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+        ACI 318 uses Load and Resistance Factor Design (LRFD), which
+        differs from the Eurocode / fib Model Code partial factor
+        method used elsewhere in this library. Two consequences for
+        users of this class:
+
+        1. Material strengths are not reduced at the material level.
+           gamma_c defaults to 1.0 and should be left at 1.0 for
+           standard ACI 318 design. Reducing fc' via gamma_c is not
+           ACI 318 compliant.
+
```
aperezcaldentey: I agree, so don't give the user the possibility of doing otherwise.

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+        Arguments:
+            fck (float): Specified compressive strength (fc') in MPa.
+
+        Keyword Arguments:
+            name (str): A descriptive name for concrete.
+            density (float): Density of material in kg/m3
+                (default: 2320).
+            gamma_c (float, optional): Partial factor for concrete.
+                Default is 1.0 (ACI does not use material partial
+                factors).
```
aperezcaldentey: So we should not give the user this possibility. Delete parameter or do we need it for compatibility reasons? If so we should somehow force the user to input a value of 1.0.

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+                'elastic', 'parabolarectangle',
+                'bilinearcompression'.
+            initial_strain (Optional[float]): Initial strain.
+            initial_stress (Optional[float]): Initial stress.
+            strain_compatibility (Optional[bool]): If True, the
+                material deforms with the geometry.
+            Ec (float, optional): The modulus of elasticity in MPa.
+            fr (float, optional): The modulus of rupture in MPa.
+            fct (float, optional): The splitting tensile strength
+                in MPa.
```
aperezcaldentey: As commented above fct is not defined in ACI 318-19. We should not use it.

structuralcodes/materials/concrete/_concreteACI318.py
```diff
@@ -0,0 +1,289 @@
...
+            wc (float): Unit weight of concrete in kg/m3
+                (default: 2320).
```
aperezcaldentey: Justify default value.

structuralcodes/materials/reinforcement/_reinforcementACI318.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @property
+    def gamma_s(self) -> float:
+        """The partial factor for reinforcement.
+
+        Note:
+            Default is 1.0 for ACI 318 (no material partial
+            factor).
+        """
+        return self._gamma_s or 1.0
```
aperezcaldentey: As commented on before for ACI we should set all material partial factors to 1.00 and not allow the user to change them. I don't think any good will come from this freedom.

structuralcodes/codes/__init__.py
gabe-kafka: this convention makes sense and you are right about the -19 -25 signaling release year. I will focus on ACI318-19 for now.

structuralcodes/materials/concrete/_concreteACI318.py
gabe-kafka: understood and fixed

structuralcodes/materials/reinforcement/_reinforcementACI318.py
gabe-kafka: done

gabe-kafka:
Pushed a rebased update against current `dev` addressing the ACI 318-19 review comments.

Summary of the changes:
- Versioned the code module and classes as `aci318_19`, `ConcreteACI318_19`, and `ReinforcementACI318_19`.
- Removed `fct`, reordered the ACI concrete functions by code sequence, used the requested exact `beta1` expression, and enforced the 17 MPa lower limit.
- Separated base material `density` from ACI equilibrium density `wc`; `Ec` and `lambda_factor` now use the ACI density path where applicable.
- Kept `gamma_c` and `gamma_s` only for shared API compatibility; non-`1.0` values now raise `ValueError`.
- Added the reinforcement grade table reference, and did not add grade-level ultimate strain because ACI 318-19 Table 20.2.1.3(a) does not provide a single value independent of reinforcement specification and bar size.
- Added customary-unit helpers plus `ConcreteACI318_19.from_psi(...)`, `ReinforcementACI318_19.from_grade(...)`, and `ReinforcementACI318_19.from_ksi(...)`, with SI storage internally.

Verification after rebasing:
- `./venv/bin/python -m ruff check structuralcodes/codes/aci318_19 structuralcodes/materials/concrete/_concreteACI318_19.py structuralcodes/materials/reinforcement/_reinforcementACI318_19.py tests/test_aci318_19` - passed
- `./venv/bin/python -m ruff format --check structuralcodes/codes/aci318_19 structuralcodes/materials/concrete/_concreteACI318_19.py structuralcodes/materials/reinforcement/_reinforcementACI318_19.py tests/test_aci318_19` - passed
- `./venv/bin/python -m pytest tests/test_aci318_19` - 84 passed
- `./venv/bin/python -m pytest` - 10,240 passed, 1 local EC2 2023 mismatch

For the EC2 2023 point: I should not have described this as a confirmed upstream/pipeline failure. It is reproducible locally in this workspace, while CI does not appear to show it. The exact local failure is `tests/test_ec2_2023/test_ec2_2023_section5_materials.py::test_eps_cs_50y[25-dry-800-CN-0.46333]`, where the actual value is `0.45333333333333337` and the expected value is `0.46333`. I am treating it as local/environment-specific unless you would like it triaged separately.

PR #344, opened 2026-04-13
fix: raise a deprecation warning if a GenericSection is created

mortenengen:
If a `GenericSection` is created, a `DeprecationWarning` is raised, and a `BeamSection` is created instead.


talledodiego (approved):
Perfect!

PR #348, opened 2026-04-20
fix: annotate section_calculator of BeamSection to improve user experience

### LINKED ISSUE #351: Type annotation error with BeamSectionCalculator (ty)
Hello,

Thanks for this really nice and impressive library. 🤗😎

I'm using VsCode with ty for auto completion/type checking.

It access to details about the BeamSectionCalculator's method by the BemSection class. But i've an  unresolved-attribute.

section.section_calculator.calculate_bending_strength()

Image

I made a try with a fork of the last updated repo. All test are passed fine.

Image

If think it's because of the to much abstraction of ABC SectionCalculator class ? BeamSectionCalculator is "child" of SectionCalculator, where they're is no calculate_bending_strength method ? 

I would like to know if it's just by my side or for everyone ? And if I can do something for ?

I had the same issue on 3 different configuration (macos/windows with UV/Vscode/venv)

It's my first public issue so I hope I'm not doing this wrong 😅

mortenengen:
After #331, the `section_calculator` attribute of the `BeamSection` is only annotated as the base class `SectionCalculator`, and the abstract methods were removed from the base class to avoid circular imports. `section_calculator` is now properly annotated as `BeamSectionCalculator` to give proper code hints in editors.

### DIFF
--- a/structuralcodes/sections/_beam_section.py
+++ b/structuralcodes/sections/_beam_section.py
@@ -44,6 +44,7 @@ class BeamSection(Section):
     """
 
     geometry: CompoundGeometry
+    section_calculator: BeamSectionCalculator
 
     def __init__(
         self,


talledodiego (approved):
Thanks!

PR #349, opened 2026-04-21
feat: improve default names

mortenengen:
## Description

This is an improvement to the naming of core components in the library. It is an attempt to implement a global counter for materials, constitutive laws, geometries and sections in a consistent manner, and use these counters as means for creating default object names when names are not given at object initialization.

## To do

- [x] Implement `base_name` argument in `Geometry` base class.
- [x] Remove `group_label` from `Geometry` base class and move to only those classes where it is relevant.

### DIFF
--- a/structuralcodes/core/base/_constitutive_law.py
+++ b/structuralcodes/core/base/_constitutive_law.py
@@ -13,13 +13,15 @@ class ConstitutiveLaw(abc.ABC):
     """Abstract base class for constitutive laws."""
 
     __materials__: t.Tuple[str] = ()
-    constitutive_law_counter: t.ClassVar[int] = 0
+    _constitutive_law_counter: t.ClassVar[int] = 0
+    id: int
 
-    def __init__(self, name: t.Optional[str] = None) -> None:
+    def __init__(
+        self, name: t.Optional[str] = None, base_name: str = 'ConstitutiveLaw'
+    ) -> None:
         """Initialize a ConstitutiveLaw object."""
-        self.id = self.constitutive_law_counter
-        self._name = name if name is not None else f'ConstitutiveLaw_{self.id}'
-        self._increase_global_counter()
+        self.id = ConstitutiveLaw.return_global_counter_and_increase()
+        self._name = name if name is not None else f'{base_name}_{self.id}'
 
     @property
     def name(self):
@@ -28,7 +30,14 @@ class ConstitutiveLaw(abc.ABC):
 
     @classmethod
     def _increase_global_counter(cls):
-        cls.constitutive_law_counter += 1
+        cls._constitutive_law_counter += 1
+
+    @classmethod
+    def return_global_counter_and_increase(cls):
+        """Returns the current counter and increases it by one."""
+        counter = cls._constitutive_law_counter
+        cls._increase_global_counter()
+        return counter
 
     @abc.abstractmethod
     def get_stress(
--- a/structuralcodes/core/base/_geometry.py
+++ b/structuralcodes/core/base/_geometry.py
@@ -11,7 +11,8 @@ from ._material import Material
 class Geometry:
     """Base class for a geometry object."""
 
-    section_counter: t.ClassVar[int] = 0
+    _geometry_counter: t.ClassVar[int] = 0
+    id: int
 
     def __init__(
         self, name: t.Optional[str] = None, group_label: t.Optional[str] = None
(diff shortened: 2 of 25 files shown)


structuralcodes/core/base/_geometry.py
```diff
@@ -11,7 +11,8 @@
 class Geometry:
     """Base class for a geometry object."""
 
-    section_counter: t.ClassVar[int] = 0
+    _geometry_counter: t.ClassVar[int] = 0
+    id: int
```
talledodiego, line 58: I have a doubt: `CompoundGeometry` is a derived class of `Geometry`. Should this have `id` `name` and/or `group_label`? 
In particular, it seems is not touched in this PR, but a design "issue" from before. But what happens is that the + creates a coupound from a list without calling the super? Therefore never defining the id, name and group_label. It actually makes sense from a logic point of view in my mind,but being derived from Geometry it should have such base attributes? I am confused...

structuralcodes/core/base/_constitutive_law.py
```diff
@@ -13,13 +13,15 @@ class ConstitutiveLaw(abc.ABC):
     """Abstract base class for constitutive laws."""
 
     __materials__: t.Tuple[str] = ()
-    constitutive_law_counter: t.ClassVar[int] = 0
+    _constitutive_law_counter: t.ClassVar[int] = 0
+    id: int
```
talledodiego, line 17: why do we need both a counter and an id? Aren't those redundant?

structuralcodes/core/base/_constitutive_law.py
```diff
@@ -13,13 +13,15 @@ class ConstitutiveLaw(abc.ABC):
...
 
-    def __init__(self, name: t.Optional[str] = None) -> None:
+    def __init__(
+        self, name: t.Optional[str] = None, base_name: str = 'ConstitutiveLaw'
```
talledodiego, line 20: why having a base_name attribute instead of coding directly the string in the else (when name is None)? The same applies to other functions similar to this.

talledodiego (changes requested):
Thanks for this PR. I have some minor questions that probably are due to me missing something.

structuralcodes/core/base/_constitutive_law.py
mortenengen: The counter keeps track of the total number of constitutive law objects, but the id is an attribute of each constitutive law object that stores the count at the time of initializing the object. I suggest to maintain both in order to be able to create default names that correspond to the number at the time of initialization.

structuralcodes/core/base/_constitutive_law.py
mortenengen: I wanted to maintain setting the default name as few places as possible and decided to put this in the base class. Derived classes should be able to override the base name, e.g. it would be convenient to be able to have names that distinguish an Elastic constitutive law from a Sargin law. My suggestion is to factor out this parameter as an argument to the constructor of the base class.

I realize now that for some reason I did not implement the `base_name` argument in the `Geometry` base class. I will fix this.

structuralcodes/core/base/_geometry.py
mortenengen: I agree with your confusion, and this was not intentional from my side. Thanks for pointing it out. I think `group_label` is not applicable for the `CompoundGeometry` since it behaves mostly as a container for other geometries. However, to be consistent with the rest of the design, I guess we should add a call to super to ensure that it gets an id and a default name. Do you agree?

structuralcodes/core/base/_constitutive_law.py
talledodiego: mmm maybe I get it: you are saying that the id is not used just for the creation of the name (for which probably the counter is enough). But you also store this id in order to be used eventually later on?

structuralcodes/core/base/_geometry.py
talledodiego: I agree on this proposal.

Further: if group_label is not a property common to all Geometries it should be probably removed from the base class and added only on those for which it has a meaning?

mortenengen: I agree 👍

structuralcodes/core/base/_constitutive_law.py
mortenengen: Exactly, that is a better formulation 😃

structuralcodes/core/base/_geometry.py
mortenengen: In 3680b72, I factored out `group_label` from the base class, and rather created a mixin that could hold the logic. Do you agree?

structuralcodes/core/base/_constitutive_law.py
mortenengen: In dbbc48a I have added the `base_name` argument to the `Geometry` base class, but kept the default as it was.

talledodiego (approved):
excellent design. I like the idea of the mixin

PR #357, opened 2026-05-21
fix: improve the speed of get_balanced_failure_strain

mortenengen:
In `.get_balanced_failure_strain` on the `BeamSectionCalculator` we perform a significant amount of queries in a double for loop. This is an attempt to make the method more efficient.


talledodiego (approved):
You are right. Those lines can be factored out from the inner loop.
I think this is a good improvement.

PR #359, opened 2026-05-29
perf: build CompoundGeometry.geom lazily

### LINKED ISSUE #358: Build the SVG representation of CompoundGeometry lazily
`CompoundGeometry.__init__` builds `self.geom`, the shapely `MultiPolygon` that `_repr_svg_` uses to display the section. It is constructed on every instantiation by buffering each point geometry into a circle. 

This `MultiPolygon` is only ever needed for the SVG representation, but it is rebuilt on every construction — including every `rotate()` and `translate()` call inside the section calculators, where the section is never displayed. For reinforced sections this adds noticeable overhead to the equilibrium calculations. 

It would be faster to build `geom` lazily, computing and caching it on first access via a property rather than in `__init__`. The behaviour on access stays the same.

 I have a change ready for this and will open a PR against `dev`, as discussed with @mortenengen.

magnusfjeldolsen:
Closes #358

`CompoundGeometry.__init__` eagerly built `self.geom` — the `MultiPolygon`
used only by `_repr_svg_` for display — by buffering every point geometry into
a circle. This ran on **every** construction, including every `rotate()` call
inside the equilibrium solvers, even though the result is only ever needed for
SVG display.

This PR defers it to a lazy `geom` property computed on first access (cached in
`_geom`). Behaviour on read is unchanged: `geom` returns the same `MultiPolygon`
and `_repr_svg_` is untouched.

### Performance (medium-complexity RC section, best of 5, vs `dev`)
- Marin integrator: bending capacity ~1.3×, moment–curvature ~1.4×, N–M ~1.4× 
(marin reconstructs geometry once per iteration)
- Fiber integrator: ~1.0× 
(fiber reconstructs geometry only once per calculation)

### Risk assessment
Low, in my assessment. No material state, no public API change. The lazy property is the
only writer of `_geom`; `_repr_svg_` is the only reader of `geom`. All
construction paths (`__init__`, `rotate`/`translate`/`mirror`/`__add__`/
`__sub__`/`from_geometry`) initialise `_geom`.

### Tests
Full geometry + section suite passes; `ruff format` and `ruff check` clean.
Added `test_compound_geometry_lazy_geom` covering lazy build, correct contents,
caching, and recomputation after a transform.

### DIFF
--- a/structuralcodes/geometry/_geometry.py
+++ b/structuralcodes/geometry/_geometry.py
@@ -824,27 +824,42 @@ class CompoundGeometry(Geometry):
                 geometries, materials
             )
             self.point_geometries = []
-            # useful for representation in svg
-            geoms_representation = [g.polygon for g in self.geometries]
-            self.geom = MultiPolygon(geoms_representation)
+            # self.geom (svg representation) is built lazily, see the geom
+            # property.
+            self._geom = None
             return
         if isinstance(geometries, list):
             self.geometries, self.point_geometries = _process_geometries_list(
                 geometries
             )
-            # useful for representation in svg
-            geoms_representation = [g.polygon for g in self.geometries]
-            geoms_representation += [
-                pg.point.buffer(pg.diameter / 2)
-                for pg in self.point_geometries
-            ]
-            self.geom = MultiPolygon(geoms_representation)
+            # self.geom (svg representation) is built lazily, see the geom
+            # property.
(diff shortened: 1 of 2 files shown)


talledodiego (approved):
I totally agree with this, thanks!

mortenengen (approved):
Thanks for this useful improvement!

PR #296, opened 2025-10-30
fix: refactoring integration data for performance improvement of marin integration

talledodiego:
This PR contains the refactoring of `triangulation_data` stored initially for fiber integration into a more general `integration_data`. For fiber integration it contains the same information derived from triangulation. For marin integration now it stores as numpy arrays the reinforcement in order to make the integration faster, especially when many fibers are added.
The performance increase is of about 25-30% in time reduction for sections with many bars.

Now that bottleneck is disappeared; inspecting carefully with profiling this code:

```
from structuralcodes.materials.concrete import create_concrete
    from structuralcodes.materials.reinforcement import create_reinforcement
    from structuralcodes import set_design_code
    from structuralcodes.geometry import RectangularGeometry, add_reinforcement_line
    from structuralcodes.sections import GenericSection

    set_design_code('mc2010')

    concrete = create_concrete(30)
    reinforcement = create_reinforcement(450, 200000, 450, 0.075)

    geo = RectangularGeometry(width=200, height=4000, material=concrete)
    for y in [-60, 60]:
        geo = add_reinforcement_line(geo, (y, -1960), coords_j=(y, 1960), diameter=10, material=reinforcement, s=70)

    section = GenericSection(geometry=geo)
    section.section_calculator.calculate_mm_interaction_domain()
 ```

The total time was about 28 s, among which the vast majority (26.8 s) spent with the 32 calls to `find_equilibrium_fixed_pivot`. Inside here there was 7.5s of processing the reinforcement (among which 5s getting stress on reinforcement).

After this fix, the same code takes 21 s (-27%), among which the vast majority (19.6) spent with the same method `find_equilibrium_fixed_pivot` called 32 times. The time for processing the reinforcement dropped significantly from 7.5s to 0.16s and get_stress dropped from 5 s to 0.1s, showing the increase in performance obtained with this PR.

PR #363, opened 2026-06-09
fix: wrong reference to integration data attribute

### LINKED ISSUE #362: bug: Changing the width of the quick start example throwing an exception
Changing the width of the quick start example is causing the moment_curvature calculation to fail with the error underneath. AI suggest this is a refactoring issue in the library itself.

## ERROR
``` bash
'BeamSectionCalculator' object has no attribute 'triangulated_data'
  File "E:\Github\Python_Testing\structural_codes\quick_start.py", line 71, in <module>
    moment_curvature = section.section_calculator.calculate_moment_curvature()
AttributeError: 'BeamSectionCalculator' object has no attribute 'triangulated_data'
```

### Root Cause

The `structuralcodes` library has an incomplete refactoring. In `BeamSectionCalculator.__init__`, the attribute was renamed:

```python
# __init__ sets the NEW name:
self.integration_data = None   # renamed from triangulated_data
```

But one of the internal helper methods (`_quick_exponential_find` / `_slow_linear_find` inside `_prefind_range_curvature_equilibrium`) still uses the **old name**:

```python
# Old code still present in the library:
self.integrator.integrate_strain_response_on_geometry(
    geom, [eps_0_b, curv, 0], tri=self.triangulated_data  # <-- AttributeError!
)
```

### Why quick_start.py works but yours doesn't

The faulty code path (`_prefind_range_curvature_equilibrium`) is only triggered when the initial neutral axis guess has an axial force imbalance `> 1e-2`. In `find_equilibrium_fixed_curvature`:

```python
if abs(dn_a) <= 1e-2:
    return [eps_0_a, curv, 0]   # quick_start.py exits here (small 250×500 section)
eps_0_b, dn_b = self._prefind_range_curvature_equilibrium(...)  # rectange_analysis.py hits this
```

The larger **1000×500 mm** section produces a larger initial imbalance, pushing it past the `1e-2` threshold and into the buggy code path. The smaller 250×500 section stays within tolerance and exits early.

## Example Code
``` python
"""Quickstart example."""

from shapely import Polygon

from structuralcodes import set_design_code
from structuralcodes.geometry import SurfaceGeometry, add_reinforcement
from structuralcodes.materials.concrete import create_concrete
from structuralcodes.materials.reinforcement import create_reinforcement
from structuralcodes.sections import BeamSection

import seaborn as sns

# Set the active design code
set_design_code('ec2_2004')

# Create a concrete and a reinforcement
fck = 45
fyk = 500
ftk = 550
Es = 200000
epsuk = 0.07

# These factory functions create concrete and reinforcement materials according
# to the globally set design code
concrete = create_concrete(fck=fck)
reinforcement = create_reinforcement(fyk=fyk, Es=Es, ftk=ftk, epsuk=epsuk)

# Create a rectangular geometry
width = 1000
height = 500
polygon = Polygon(
    [
        (-width / 2, -height / 2),
        (width / 2, -height / 2),
        (width / 2, height / 2),
        (-width / 2, height / 2),
    ]
)  # We leverage shapely to create geometries
geometry = SurfaceGeometry(
    poly=polygon, material=concrete
)  # A SurfaceGeometry is a shapely Polygon with an assigned material

# Add reinforcement
diameter_reinf = 25
cover = 50

geometry = add_reinforcement(
    geometry,
    (
        -width / 2 + cover + diameter_reinf / 2,
        -height / 2 + cover + diameter_reinf / 2,
    ),
    diameter_reinf,
    reinforcement,
)  # The add_reinforcement function returns a CompoundGeometry
geometry = add_reinforcement(
    geometry,
    (
        width / 2 - cover - diameter_reinf / 2,
        -height / 2 + cover + diameter_reinf / 2,
    ),
    diameter_reinf,
    reinforcement,
)

# Create section
section = BeamSection(geometry)

# Calculate the moment-curvature response
moment_curvature = section.section_calculator.calculate_moment_curvature()
```

mortenengen:
Two references to the attribute `triangulated_data` that were missed in #296 are now corrected.

### DIFF
--- a/structuralcodes/sections/_beam_section.py
+++ b/structuralcodes/sections/_beam_section.py
@@ -559,7 +559,7 @@ class BeamSectionCalculator(SectionCalculator):
                 _,
                 _,
             ) = self.integrator.integrate_strain_response_on_geometry(
-                geom, [eps_0_b, curv, 0], tri=self.triangulated_data
+                geom, [eps_0_b, curv, 0], tri=self.integration_data
             )
             dn_b = n_int - n
             if dn_a * dn_b < 0:
@@ -595,7 +595,7 @@ class BeamSectionCalculator(SectionCalculator):
                 ) = self.integrator.integrate_strain_response_on_geometry(
                     geom,
                     [eps_0_attempts[j], curv, 0],
-                    tri=self.triangulated_data,
+                    tri=self.integration_data,
                 )
                 dn_attempts[j] = n_int - n
             if dn_a > 0:
--- a/tests/test_sections/test_beam_section.py
+++ b/tests/test_sections/test_beam_section.py
@@ -2019,3 +2019,59 @@ def test_perimeter_multipolygon():
         gp = section.gross_properties
 
     assert math.isclose(gp.perimeter, 0)
+
+
+def test_wide_section():
+    """Test calculating the moment-curvature relation for a wide section.
+
+    Regression test for #362. Based on the geometry from the quickstart example
+    with an increased width.
+    """
+    # Create a concrete and a reinforcement
+    fck = 45
+    fyk = 500
+    ftk = 550
+    Es = 200000
+    epsuk = 0.07
+
+    concrete = ConcreteEC2_2004(fck=fck)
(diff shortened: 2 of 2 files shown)


talledodiego (approved):
Thanks for having solved this bug!

PR #367, opened 2026-07-29
Fix Python compatibility metadata for 3.14 releases

Piripicchio66:
## Summary
- Relax the declared Python compatibility range to allow Python 3.14 patch releases such as 3.14.4.
- Bump the package version to 0.7.2 so the release workflow can publish a new distribution.
- Add a regression test covering the Python requirement metadata.

## Why
The current metadata used `>=3.10,<=3.14`, which excludes Python 3.14.4 and causes pip to reject installation with a dependency conflict.

## Verification
- `py -m pytest -q tests/test_package_metadata.py` ✅
- `py -m build` ✅


mortenengen:
Thanks for requesting this, @Piripicchio66. The reason for the rather strict compatibility range is the required dependency `triangle` which currently only supports python<=v3.13. We are looking for alternatives that are actively maintained to avoid this unwanted lag in Python version compatibility. Until then, the proposed changes unfortunately do not solve the issue, and I am therefore closing the issue.

PR #369, opened 2026-08-28
fix: fix marin integration for more than one reinforcement material

### LINKED ISSUE #368: bug: marin integration fails when more than one material is present for point geometries
Hello
I have noticed that the Marin integrator fails when two or more types of reinforcing steel are used. The error appears to lie in the “_process_point_geometries” function.
Thanks!

mortenengen:
The Marin integrator fails when more than one reinforcement material is present and the number of points for each material is different. This is an attempt to fix this behavior.

### DIFF
--- a/structuralcodes/sections/section_integrators/_marin_integrator.py
+++ b/structuralcodes/sections/section_integrators/_marin_integrator.py
@@ -223,9 +223,6 @@ class MarinIntegrator(SectionIntegrator):
             )
         # Now use the integration data (either just created or passed as
         # argument) to fill the input
-        y = []
-        z = []
-        IA = []
         for reinf_data in integration_data:
             # All have the same constitutive law
             strains = strain[0] + strain[1] * reinf_data[1]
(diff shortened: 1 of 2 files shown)


structuralcodes/sections/section_integrators/_marin_integrator.py
```diff
@@ -240,7 +240,15 @@ def _process_point_geometries(
             z.append(reinf_data[1])
             IA.append(integrand * reinf_data[2])
 
-        input.append((1, np.array(y), np.array(z), np.array(IA)))
```
marianolopezlema, line 243: y = np.asarray(reinf_data[0])
            z = np.asarray(reinf_data[1])
            IA = np.asarray(integrand * reinf_data[2])

            input.append((1, y, z, IA))

mortenengen: This is a good observation. I have simplified the processing in 879c0cd.

talledodiego (approved):
Fine with me, thanks for having spotted and solved this.

PR #372, opened 2026-09-07
fix: change default convergence tolerance for beam section calculations

mortenengen:
In some situations, the default convergence tolerance was not enough to reach satisfying equilibrium. A regression test is added and the default convergence tolerance is now slightly tightened.


talledodiego (approved):
Thanks for this!

PR #376, opened 2026-09-18
feat: Add elastic fragile tension constitutive law for concrete

Pier53:
Hello everyone,

Please be lenient, as this is my first real public pull request. I hope I haven't made a mistake! :-)

I've started to implement ElasticFragileTension.

- [x] Add a _basic_ class ElasticFragileTension(ConstitutiveLaw)
- [x] .get_ultimate_strain() (chapter bellow)
- [x] Add __marin__() method
- [x] Add __marin_tangent__() method
- [ ] Add tests
- [ ] Other ?

### .get_ultimate_strain()
I had a trouble with .get_ultimate_strain() with the Parallel class because compression law returned epsillon_tension_max = 100 instead of 0.00.

```python
compression = ParabolaRectangle(fc=45)      # ou Sargin, Popovics...
tension = ElasticFragileTension(fct=2.7, Ec=36000)
compression_tension = Parallel(constitutive_laws=[compression, tension])

compression_tension.get_ultimate_strain()
```
return :
[
  -0.0035,
  **100**
]

So the ultimate strains of the compression_tension made with Parallel class was wrong. 

__| compression | tension | compression_tension
-- | -- | -- | --
classe | ParabolaRectangle | ElasticFragileTension | Parallel
negative | -3.500 ‰ | 0.000 ‰ | -3.500 ‰
positive | 100000.000 ‰ | 0.075 ‰ | 100000.000 ‰

Maybe you've already talk about it in past ?
I've thinked of different solutions.

1. Replacing by None --> Wrong because we are going to insert None instead of float, and it's not clear to understand what it's mean to have a None Ultimate Strain...
2. Replacing by 0.00 --> Wrong idea (break constitutive laws when they are alone)
3. Replacing 100.0 by np.inf (or -np.inf)
4. Replacing by NaN --> It's like np.inf

So I've made a try with np.inf !

```python
#--- Compression law ---
def get_ultimate_strain(self, **kwargs) -> t.Tuple[float, float]:
    """Return the ultimate strain (negative and positive).
    """
    del kwargs
    return (self._eps_cu, math.inf)


#--- Parallel law ---
def get_ultimate_strain(self, yielding: bool = False) -> t.Tuple[float, float]:
    """Return the ultimate strain (negative and positive)."""
    neg = [law.get_ultimate_strain(yielding=yielding)[0] for law in self._wrapped_laws]
    pos = [law.get_ultimate_strain(yielding=yielding)[1] for law in self._wrapped_laws]

    finite_neg = [v for v in neg if np.isfinite(v)]
    finite_pos = [v for v in pos if np.isfinite(v)]

    eps_n = min(finite_neg) if finite_neg else -np.inf
    eps_p = max(finite_pos) if finite_pos else np.inf
    return (eps_n, eps_p)
```

With this, i got good results :

__| compression | tension | compression_tension
-- | -- | -- | --
classe | ParabolaRectangle | ElasticFragileTension | Parallel
negative | -3.500 ‰ | -inf ‰ | -3.500 ‰
positive | inf ‰ | 0.075 ‰ | 0.075 ‰

But this was introducing some trouble in BeamSectionCalculator so i introduce a "bracket" with a large strain of 1.0.

My test file : [2026_09_18_bending_strength_concrete_only.py](https://github.com/user-attachments/files/32389982/2026_09_18_bending_strength_concrete_only.py)

### Marin
I add a minimum for eps to avoid to get an np.inf value in the discretize_law
```python
eps_min, eps_max = self.get_ultimate_strain()
eps_min = max(eps_min, -1)
eps_max = min(eps_max, 1)
```

-> Do i need to add a specific marin method for ElasticFragileTension ?

### Tests
All test passed.

### Roadmap

- Waiting your review and applying it
- Add tests
- Next ?

I reallyenjoy doing this, but I'm not used to work on project on Github, so I need your advises ! 👍


mortenengen:
Thanks for this contribution, @Pier53! I am sorry that you had to run into that issue with the strain limits. Indeed, I remember @talledodiego and I discussed what to do with the limits, and we decided to keep it like this awaiting a tensile law 😃 Thanks for elaborating on this in the message above.

Anyway, I will have a closer look after the weekend, and @talledodiego will have a look at your solution for the strain limit issue.

If possible, please go ahead implementing the `__marin__` and the `__marin_tangent__` methods for completeness.

structuralcodes/core/base/_constitutive_law.py
```diff
@@ -124,8 +125,8 @@ def find_x_lim(x, y):
 
         eps = np.concatenate((eps_neg, eps_pos))
         sig = self.get_stress(eps)
-        from structuralcodes.materials.constitutive_laws import (  # noqa: PLC0415
-            UserDefined,
+        from structuralcodes.materials.constitutive_laws import (
```
mortenengen, line 128: I am not sure why this change is suggested, but anyhow, the result is that the `# noqa: PLC0415` flag to `ruff` is not captured, and the pipeline does not pass. Please add a commit where this change is undone such that the pipeline passes.

mortenengen (changes requested):
Thanks for this nice contribution, @Pier53! I think the use of `np.inf` is a clever solution. I only have one comment which I would appreciate if you could fix in order to trigger the pipelines. Final approval awaits the outcome of the pipelines, i.e. whether they pass or not, and the review from @talledodiego.

Pier53:
I think I made a mistake with the get_ultimate_strain method. Because my result with and without rebar are inconsistent/wrong. Please wait a bit before the review because I'm going to fix it soon.
But I'm not totally available this week. Thanks 

I'm also going to fix this noqa bug.

